PR #9, opened 2020-09-21
Upload pbtxt dataset for Merck CN cross coupling

beef-broccoli:
This is enumerated dataset of 96 reactions from experiment 1 of paper 10.1126/science.1259203. Created with web editor. Template reaction pbtxt and csv are attached. 

Potential problems with this set: most nanomole scale reactions in the SI seem to only have relative conversion, no actual yield (Internal standard was used in UPLC analysis, but no amount; paper also did not mention what the internal standard is; only has a product/internal standard ratio). In the paper, relative conversion was compared across plate, promising conditions then scaled up to get isolated yield. I assumed that the same condition at different scales give the same yield, and based on the scaled-up isolated yield and relative conversion ratio at nanomole scale I calculated an approximate yield for all nanomole reactions

[template.zip](https://github.com/Open-Reaction-Database/ord-data/files/5258325/template.zip)


beef-broccoli:
I only did experiment 1 because this yield issue, but can definitely work on other data in the SI after we figure out what's the best way to deal with it

connorcoley:
Thanks for being patient! It's taken me longer to get to this than I would have liked. We're still working on the review process, so please treat this as a two-way conversation!

You've run into some tricky questions about recording yield that we've been discussing internally as well. I've opened an [issue](https://github.com/Open-Reaction-Database/ord-schema/issues/432) to discuss how to best capture this.

Notes:
- Confirmed amounts of two reactants (in common for all experiments), amount of base in DMSO (constant amount, but identity varies), and amount of Pd precatalyst in DMSO (constant amount, but identity changes)
- Experiment 1 actually contains two different experiments: one that is smaller scale, performed in a 1536 well plate, and one that is larger scale, performed in 96 well plates. Both are recorded in Data S1 it seems. Based on the quantities in your template, I am assuming you are intending to capture the 1536 well plate version.

Major suggestions:
- The quench appears to be described as 3 microliters of the acetic acid + biphenyl solution in DMSO; however, the template records the amount as 10 mL. Was this intentional?
- I am going to create a new workup step that is of type `ALIQUOT` that will let us capture the fact that 1 microliter of the quenched solution was sampled, and that 75 more microliters of DMSO was added (dilution). Once that is merged into the schema, I would suggest we add two workup steps: ALIQUOT and then ADDITION.
- I'm going to contradict myself from the video tutorial, but if the paper does not actually specify what room temperature is, it's better to leave the control type as AMBIENT and _not_ guess a setpoint. Sorry for creating confusion.
- I am reasonably confident that biphenyl is the internal standard they use for quantification. It isn't explicitly stated in the section on Experiment 1, but the other internal standards for other experiments are biphenyl derivatives.
- Once we figure out how best to capture the fact that the yields are all relative, it will still be worth elaborating on how you calculated the yields -- Perhaps that should be included in the `details` string of the analysis?
- Since these 96 experiments were actually run in triplicate, I think it would be worth including each of the individual outcomes as separate entries in the database rather than using the average performance of the three. Would that make sense to you?

Minor suggestions:
- Were all of the SMILES strings for the Pd catalysts obtained through name resolution? If not, I'd suggest removing the details string of "NAME resolved by the PubChem API" from the template, since that will get recorded for every structure.
- It could be valuable to specify the automation platform as "Mosquito (for liquid handling)"
- Although Tim and Spencer were the corresponding authors, we don't know who the experimenter actually was. In this case, I think providing the link to the publication (and adding the DOI to the DOI field) is sufficient.

skearnes:
Hi @beef-broccoli, have you had a chance to review the comments posted by @connorcoley?

beef-broccoli:
@connorcoley @skearnes Sorry for the delay. The suggestions make sense, I'll try uploading an updated version this week

skearnes:
Hi @beef-broccoli, just checking in. We're going to be making some changes to the schema soon (see https://github.com/Open-Reaction-Database/ord-schema/pull/496) and it would be nice to have this merged before then so you don't have to change anything.

beef-broccoli:
@skearnes I apologize, things have been busy. I will update them today according to suggestions

beef-broccoli:
Checks have failed, but I dont think that's me. The template is validated in the web editor before enum. The dataset after enum does give me a validation error in the web editor in the workup section: the internal standard biphenyl was using "volume" as "amount" and "nanomole" as unit. But i checked the template as well as the dataset, it is correct with "moles" as unit of amount. The checks output indicates it's some parsing error with "identifiers", not sure what happened. 

Template and csv files are attached



[Merck_cross_coupling_exp1_v3.zip](https://github.com/Open-Reaction-Database/ord-data/files/5522924/Merck_cross_coupling_exp1_v3.zip)

beef-broccoli:
Is it because my forked repo is too old? Did not sync since I forked. Pretty confident the template is correct

beef-broccoli:
I think all the errors are resolved. Let me know if there is anything else

skearnes:
Thanks! We'll take a look ASAP.

connorcoley:
Thanks for these changes! The very last thing I'd suggest you modify relates to the workup. I noticed you added a volume of 1 microliter to the first ADDITION workup step that includes the biphenyl solution. Since we've made ALIQUOT its own workup step type, I'd suggest that this reaction actually needs a total of three workup steps: ADDITION, ALIQUOT, then ADDITION.

The zipped pbtxt here is a slightly modified template with only that changed: 
[Merck_cross_coupling_exp1_v3_cwc.pbtxt.zip](https://github.com/Open-Reaction-Database/ord-data/files/5534373/Merck_cross_coupling_exp1_v3_cwc.pbtxt.zip)

After this tiny change, I think we're good to go!

beef-broccoli:
And here's the template and dataset again. Again, sorry for all the delay. Really appreciate all the help from everyone!
[Merck_cross_coupling_exp1_v4.zip](https://github.com/Open-Reaction-Database/ord-data/files/5557457/Merck_cross_coupling_exp1_v4.zip)

connorcoley (approved):

PR #8, opened 2020-09-16
Updating dataset with Suzuki informer library

jun-li-2020:
[Pd_CN_Coupling_template_spreadsheet.zip](https://github.com/Open-Reaction-Database/ord-data/files/5253155/Pd_CN_Coupling_template_spreadsheet.zip)
[Suzuki _template_spreadsheet.zip](https://github.com/Open-Reaction-Database/ord-data/files/5253139/Suzuki._template_spreadsheet.zip)

Uploaded the Suzuki template and 120-rxn spreadsheet csv file.
Kutchukian et al "Chemistry informer libraries: a chemoinformatics enabled approach to evaluate and advance synthetic methods"
Ref: Chem. Sci., 2016, 7, 2604
doi: 10.1039/c5sc04751j

PR #10, opened 2020-09-27
Add a photoredox dataset from 10.1021/acscatal.0c02247

jun-li-2020:
Besides adding the two finalized Suzuki and CN informer library dataset.  I have added a new photoredox library dataset.  Detail is shown as below:

This is a 1,152-rxn photoredox HTE screening study of a photodehalogenation reaction using various iridium(III) complex photocatalysts.  Ref: Mdluli, V. et al ACS Catal. 2020, 10, 6977−6987. DOI: 10.1021/acscatal.0c02247. High-throughput Synthesis and Screening of Iridium(III)Photocatalysts for the Fast and Chemoselective Dehalogenation ofAryl Bromides.  There are a few challenges observed during the reaction template preparation.  1) The combinatorial photocatalyst library prep contains multiple steps from [Ir(C^N)2Cl]2 dimers which is not documented, to formation of [Ir(C^N)2(N^N)]+.  In order to show the combinatorial set for both CN and NN ligands, I did not use crude product function, this is similar to the dealing with previous precatalyst description. I only focus on the ligand SMILES and put detailed prep info under add preparation section. 2) This library is monitored using an unconventional colorimetric analysis with indicator dye to trap the dehalogened byproduct HBr. Also the targeted measurement is rate constant, not yield or selectivity.  I used selectivity cell to add the rate constant result and use detail to clarify it.  Since the indicator is the limiting reactant used to capture the 1st-order decay rate along with 8 different concentrations span from 0-1.4 mM across 8 rows per photocatalyst column on the 96-welll, based on the input requirement, I improvised by using the average mM with +/- range and detailed in the add prep section.  The enumeration process looked pretty good.  I noticed one error which I did not see in the previous Suzuki and CN cases.  The volume unit shown in the input material as well as setup section does NOT match the generated template or graph, for example, the vial size in the web template was "mL", while the downloaded pbttext file showed "Liter" instead.  I corrected manually to mL, when I enumerated, the web version showed Liter.  The same happened in the input solvent volume section.  I think overall this tool shows its flexibility to capture this very intricate HTE design and experiment setup.  Please review and let me know what do you guys think.
[Photodehalogen_Merged_dataset.zip](https://github.com/Open-Reaction-Database/ord-data/files/5288340/Photodehalogen_Merged_dataset.zip)


skearnes:
Hi @jun-li-2020, what is the difference between this PR and #8? I see the same three files in each. In general each publication should have its own PR.

skearnes:
I'd suggest updating the files in #8 with the "final" ones you have here and making this PR specific to the new photoredox dataset (and removing it from #8).

jun-li-2020:
> Hi @jun-li-2020, what is the difference between this PR and #8? I see the same three files in each. In general each publication should have its own PR.

Sorry for previous error I made by including all the #8 and this PR together.  To clean that up according to your suggestions. I deleted the new photodehalogenation dataset in #8 and replaced the old Suzuki/CN with finalized versions. For this PR, I deleted the Suzui/CN datasets and kept the new dataset only.

skearnes:
Thanks @jun-li-2020! A few comments:

> There are a few challenges observed during the reaction template preparation.
> 1) The combinatorial photocatalyst library prep contains multiple steps from [Ir(C^N)2Cl]2 dimers which is not documented, to formation of [Ir(C^N)2(N^N)]+. In order to show the combinatorial set for both CN and NN ligands, I did not use crude product function, this is similar to the dealing with previous precatalyst description. I only focus on the ligand SMILES and put detailed prep info under add preparation section. 

In this case I think I would separate the [Ir(C^N)2Cl]2 and N^N ancillary ligand into separate inputs (see the "Colorimetric Screening" section of the paper). The iridium dimers would ideally be drawn as SMILES with [dative bonds](https://www.rdkit.org/docs/RDKit_Book.html#dative-bonds), or less ideally as `.`-joined SMILES (including the iridium) in the same component.

> 2) This library is monitored using an unconventional colorimetric analysis with indicator dye to trap the dehalogened byproduct HBr. Also the targeted measurement is rate constant, not yield or selectivity. I used selectivity cell to add the rate constant result and use detail to clarify it. Since the indicator is the limiting reactant used to capture the 1st-order decay rate along with 8 different concentrations span from 0-1.4 mM across 8 rows per photocatalyst column on the 96-welll, based on the input requirement, I improvised by using the average mM with +/- range and detailed in the add prep section. 

Using `selectivity` is probably the best place to put the rate constants in the current schema; @connorcoley should we add an arbitrary `Data` message to `ReactionProduct`?

For the indicator dye, I think I'd prefer that you choose a single value to use for all of the reactions. Please still put the details in to indicate that the rate constant was obtained by comparing many concentrations.

> The enumeration process looked pretty good. I noticed one error which I did not see in the previous Suzuki and CN cases. The volume unit shown in the input material as well as setup section does NOT match the generated template or graph, for example, the vial size in the web template was "mL", while the downloaded pbttext file showed "Liter" instead. I corrected manually to mL, when I enumerated, the web version showed Liter. The same happened in the input solvent volume section.

Thanks for reporting this bug; I believe this was fixed in https://github.com/Open-Reaction-Database/ord-schema/pull/434 but please let us know if you see it again.

jun-li-2020:
Thank you Steve for your comments on the photocatalyst screening library.  You are probably right that the iridium bpy/ppy based dimers would ideally be drawn as SMILES with dative bonds instead of concatenated SMILES.  I know the organometallic complex representation is still an area of lots of challenges, especially having seen the recent presentation from Prof. Jan Jensen at RDKit UGM.  I haven't used the online tool to draw those complex before.  Will look into it.

skearnes:
Thanks Jun. You might want to hold off for a moment; @connorcoley, @michaelmaser, and I are discussing more about how to capture the endpoints for this assay and whether we want to adjust the schema, etc.

connorcoley:
This submission is very intriguing to me, and is making me think about the type of data that the schema is designed to collect. 

My opinion is that a Reaction message should _not_ contain data that was obtained from multiple distinct experiments. Recording the rate constant with a single Reaction forces us to define a single set of conditions (conc, time) rather than the full array of conditions that were used. I view the rate constants as a property calculated from multiple Reactions or time points, which means that they don't really belong in a single ReactionOutcome.

For this example, @jun-li-2020 , would you be willing to contact the corresponding author to see if they would be willing to share the full dataset prior to the calculation of rate constants?

jun-li-2020:
Thanks Connor, I will reach out to the authors.

skearnes:
Superseded by #40

PR #12, opened 2020-09-29
Updating dataset with Suzuki informer library (#8)

skearnes:
* Updating dataset with Suzuki informer library

Kutchukian et al "Chemistry informer libraries: a chemoinformatics enabled approach to evaluate and advance synthetic methods"
Ref: Chem. Sci., 2016, 7, 2604
doi: 10.1039/c5sc04751j

* Updating dataset with Pd CN informer library

Here uploaded is an enumerated pbtxt containing Pd catalyzed C-N coupling reactions informer library extracted from Figure_4D palladium catalyst/ligand section.  A total of of 144 reactions covering the entries from 1 to 8 were constructed using the webtool from the reference: Kutchukian et al "Chemistry informer libraries: a chemoinformatics enabled approach to evaluate and advance synthetic methods" Chem. Sci., 2016, 7, 2604. (doi: 10.1039/c5sc04751j)

* Add files via upload

This is a 1,152-rxn photoredox HTE screening study of a photodehalogenation reaction using various iridium(III) complex photocatalysts.  Ref: Mdluli, V. et al ACS Catal. 2020, 10, 6977−6987. DOI: 10.1021/acscatal.0c02247. High-throughput Synthesis and Screening of Iridium(III)Photocatalysts for the Fast and Chemoselective Dehalogenation ofAryl Bromides.  There are a few challenges observed during the reaction template preparation.  1) The combinatorial photocatalyst library prep contains multiple steps from [Ir(C^N)2Cl]2 dimers which is not documented, to formation of [Ir(C^N)2(N^N)]+.  In order to show the combinatorial set for both CN and NN ligands, I did not use crude product function, this is similar to the dealing with previous precatalyst description. I only focus on the ligand SMILES and put detailed prep info under add preparation section. 2) This library is monitored using an unconventional colorimetric analysis with indicator dye to trap the dehalogened byproduct HBr. Also the targeted measurement is rate constant, not yield or selectivity.  I used selectivity cell to add the rate constant result and use detail to clarify it.  Since the indicator is the limiting reactant used to capture the 1st-order decay rate along with 8 different concentrations span from 0-1.4 mM across 8 rows per photocatalyst column on the 96-welll, based on the input requirement, I improvised by using the average mM with +/- range and detailed in the add prep section.  The enumeration process looked pretty good.  I noticed one error which I did not see in the previous Suzuki and CN cases.  The volume unit shown in the input material as well as setup section does NOT match the generated template or graph, for example, the vial size in the web template was "mL", while the downloaded pbttext file showed "Liter" instead.  I corrected manually to mL, when I enumerated, the web version showed Liter.  The same happened in the input solvent volume section.  I think overall this tool shows its flexibility to capture this very intricate HTE design and experiment setup.  Please review and let me know what do you guys think.

* Delete Photodehalogen_Merged_dataset_dataset.pbtxt

* Delete enumerate_Pd_CN_dataset.pbtxt

* Delete enumerate_Suzuki_dataset.pbtxt

* Add files via upload

* Add files via upload

* Delete enumerate_CN_template_spreadsheet.zip

* Delete enumerate_Suzuki_template_spreadsheet.zip


connorcoley (approved):
Note for posterity: long description of additional photochemistry dataset in the PR description is not actually relevant to this PR

PR #14, opened 2020-10-05
Add Ahneman dataset

skearnes:
Forked from https://github.com/Open-Reaction-Database/ord-submissions-test/pull/27

> Review comments:
> 
> * There are ~50 fewer reactions in your csv than reported in the paper. Do you know what is causing that discrepancy? Is that the set of reactions w/o an aryl halide? In general, I would be curious to know how you prepared the csv.

The issue was the exclusion of some controls and all reactions with additive **7**. I corrected these omissions in https://github.com/doylelab/rxnpredict/pull/3. Still, after excluding some N/A rows (9) and reactions without an aryl halide (287; all zero yield obviously), we end up with 4312 reactions (out of a possible 4608).

> * It would be better to move the input corresponding to the addition of di-tert-butylbiphenyl to a workup step, rather than a reaction input. I realize that the `addition_time` does indicate that it was added after 16 hours, but an addition event immediately preceding analysis seems more like a workup input than a reaction input. This is the convention I followed for the Nielsen deoxyfluorination example for what it's worth.

Added as an `ADDITION` workup. The SI mentions that aliquots were taking for UPLC analysis, but doesn't give an amount, so I left this out---should it be included?

> * Under setup, it would be nice to specify `automation_platform = "Mosquito HTS liquid handling robot (TTP Labtech, 4.5 mm pitch tip spool)"` rather than only mentioning the Mosquito in the procedure details.

Done.

> * Under vessel, would it make sense to specify the vessel attachments? We do not have types that are suited to well-plates, which might be an oversight. It seems like we could specify two or three attachments: "perfluoroalkoxy alkane (PFA) mat (Analytical Sales, Cat. No. 96981)", "silicon rubber mat (Analytical Sales, Cat. No. 96982)", and "clamp (Arctic White, Cat. No. AWSC-051001)". I see this is included under the pressure conditions, which is much better than excluding this information. But should we add `MAT` and `CLAMP` as vessel attachment types?

Added as `CUSTOM` attachments (how similar are `MAT` and `CAP`?). If these are common attachments we could certainly add them to the enum.

> * _The product compound SMILES does not appear to be templated at the moment. Different aryl halides yield different products when they differ by more than the halogen atom. The 5 distinct product SMILES will need to be defined in a new column in your csv._

Fixed.

> * The analytical instrument should have type `LCMS`, not `LC`.

I'm not sure this is right. AFAICT the yield quantification was done chromatographically and not with MS. See Figure S2, Figure S3, and Table S3.

> 
> Picky comments:
> 
> * I'd consider the vessel material to be PLASTIC with details of "Cyclic Olefin Copolymer (COC)" rather than CUSTOM.

Done.

> * You could specify `vendor_source = "Aldrich"` for the DMSO components.

Done.

> * You could specify details of the reaction environment as "MBraun glovebox operating with a constant N2 purge (oxygen typically < 5 ppm)".

Done (and chose `NITROGEN` for the atmosphere).


skearnes:
@b-shields FYI

skearnes:
Attaching the enumeration script...

[ahneman.zip](https://github.com/Open-Reaction-Database/ord-data/files/5335083/ahneman.zip)

connorcoley:
> The SI mentions that aliquots were taking for UPLC analysis, but doesn't give an amount, so I left this out---should it be included?

If there's no amount given, then I don't think so. The default assumption wouldn't be that the entire crude mixture was injected onto the column, but some modest 1-5 microliter sample (if analytical scale).

> I'm not sure this is right. AFAICT the yield quantification was done chromatographically and not with MS. 

I guess this raises a minor question, which is when we have these combined analyses (LC + MS, which is super common) but only one is used for quantification, do we separate them or keep them together? I would assume that quantification "through LCMS" means through peak areas, but that's not unambiguous. I think your approach  makes sense here. 

Comments on template pbtxt:
- Thanks for moving the internal standard to a workup step! I might suggest you remove the `addition_order` and `addition_time` fields. These aren't relevant for workup additions, since workup steps are inherently sequential and there is only one allowed input. Perhaps we _should_ create validation checks for extraneous information in workup steps like that.

skearnes:
Updated the template to remove `addition_order` and `addition_time`. Thanks!

Updated enumeration files: [ahneman.zip](https://github.com/Open-Reaction-Database/ord-data/files/5352230/ahneman.zip)

connorcoley (dismissed):

skearnes:
Question: I modified the template/script to canonicalize the SMILES and added a component for the catalyst (we only had the ligand). PTAL; I wasn't sure whether to include a dummy atom for `L` so I left it off completely.

[ahneman.zip](https://github.com/Open-Reaction-Database/ord-data/files/5360240/ahneman.zip)

connorcoley:
Ah I didn't check for that. I think it would be preferable to draw the full catalyst/ligand complex for each one. The catalysts were purchased and used as a single component to prepare stock solutions.

I might suggest something like the following for the template:

SMILES value: `[Pd].O=S(=O)(O)C(F)(F)F.Nc2ccccc2-c2ccccc2.$ligand_smiles$`

@michaelmaser thoughts?

michaelmaser:
> I might suggest something like the following for the template since handling dative bonds is tricky:
> 
> SMILES value: `[Pd].O=S(=O)(O)C(F)(F)F.Nc2ccccc2-c2ccccc2.$ligand_smiles$`
> 
> @michaelmaser thoughts?

I think that's fine, but AFAIK this representation gets interpreted as a salt mixture or as distinct components (so 'Pd' + 'TfOH' + 'amino-biphenyl' + 'ligand'). Depending on the ligand, I've explicitly drawn out the dative bonds and generated bound complex SMILES, which gives single component representations. Example for Xphos Pd G3 is `CC(C)c1cc(C(C)C)c(-c2ccccc2[P+](C2CCCCC2)(C2CCCCC2)[Pd-2]2(OS(C)(=O)=O)[NH2+]c3ccccc3-c3ccccc32)c(C(C)C)c1` (after canonicalizing). This gives a single bound complex, though it reads the N, P, and Pd as charged to satisfy valence rules.  

Thinking out loud:
I'm not sure there's a perfect way to represent these (as SMILES at least), but from a downstream application standpoint I think it may be better to have a bound complex if trying to learn catalyst structures, for example. If splitting a reaction SMILES on `'.'`, the component representations get separated into different molecules. E.g., here, it may look like triflic acid  `O=S(=O)(O)C(F)(F)F` is being used in the reaction, though it's definitely not. It may be easier to strip charges from the bound complex as a single molecule SMILES than re-assemble the right complex from the components.

Couple docs attached with drawings and RDKit processing examples (think can only attach images/pdfs here, will send you the files).

Hope this helps. 
[pd_complex_smiles_tests_cd.pdf](https://github.com/Open-Reaction-Database/ord-data/files/5361347/pd_complex_smiles_tests_cd.pdf)
[pd_complex_smiles_tests_ipynb.pdf](https://github.com/Open-Reaction-Database/ord-data/files/5361348/pd_complex_smiles_tests_ipynb.pdf)

connorcoley:
We could use the recently-added support for dative bonds in RDKit?

`CS(O[Pd]<-1(<-[P](C2CCCCC2)(C3CCCCC3)C4=C(C5=C(C(C)C)C=C(C(C)C)C=C5C(C)C)C=CC=C4)C6=CC=CC=C6C7=C([NH2]1)C=CC=C7)(=O)=O` parses correctly

image

michaelmaser:
Nice, that's great. Saw this when looking at ferrocenes but haven't tried it yet. Is there a way to generate these automatically or did you write in the dative bonds by hand?

As for the question, I agree with including full metal/ligand complexes with each reaction.

connorcoley:
I did the dative bonds by hand. It looks like Ketcher doesn't support them when drawing or when parsing from RDKit. RDKit's MolFile has an undefined bond type that breaks Ketcher

skearnes:
This looks great; I'll push an update that uses the dative bonds. Ketcher looks to be reviving its support (tons of new issues opened in the last month or two) so maybe we can get some support there eventually.

skearnes:
PTAL and please double check my SMILES :)

[ahneman.zip](https://github.com/Open-Reaction-Database/ord-data/files/5366264/ahneman.zip)

skearnes:
@brilee PTAL?

skearnes:
@b-shields question for you: for the reactions without an additive, was the total reaction volume still 1 uL or was it only 800 uL? If the former I'll need to update the enumeration to add empty solvent when the additive is undefined.

b-shields:
My guess would be that they included a DMSO vial for the Mosquito to draw from for the additive free experiments. However, this is left ambiguous in the SI, so I am not certain.

connorcoley:
Based on the Jupyter notebook you attached, 
```
catalyst_smiles = {
    'XPhos': 'CC(C)c1cc(C(C)C)c(-c2ccccc2P(C2CCCCC2)(C2CCCCC2)->[Pd]2(OS(C)(=O)=O)<-Nc3ccccc3-c3ccccc32)c(C(C)C)c1',
    't-BuXPhos': 'CC(C)c1cc(C(C)C)c(-c2ccccc2P(C(C)(C)C)(C(C)(C)C)->[Pd]2(OS(C)(=O)=O)<-Nc3ccccc3-c3ccccc32)c(C(C)C)c1',
    't-BuBrettPhos': 'CC(C)c1cc(C(C)C)c(-c2c(OC)ccc(OC)c2P(C(C)(C)C)(C(C)(C)C)->[Pd]2(OS(C)(=O)=O)<-Nc3ccccc3-c3ccccc32)c(C(C)C)c1',
    'AdBrettPhos': 'CC(C)c1cc(C(C)C)c(-c2ccccc2P(C23CC4CC(CC(C4)C2)C3)(C23CC4CC(CC(C4)C2)C3)->[Pd]2(OS(C)(=O)=O)<-Nc3ccccc3-c3ccccc32)c(C(C)C)c1',
}
```

It looks like the only mistake is that your AdBrettPhos ligand is missing its two methoxy groups. I believe the catalyst SMILES should be `CC(C)c1cc(C(C)C)c(-c2c(OC)ccc(OC)c2P(C23CC4CC(CC(C4)C2)C3)(C23CC4CC(CC(C4)C2)C3)->[Pd]2(OS(C)(=O)=O)<-Nc3ccccc3-c3ccccc32)c(C(C)C)c1`

skearnes:
Ah good catch; fixed.

[ahneman.zip](https://github.com/Open-Reaction-Database/ord-data/files/5386480/ahneman.zip)

skearnes:
Fixed the triflates; 
[ahneman.zip](https://github.com/Open-Reaction-Database/ord-data/files/5386539/ahneman.zip)

connorcoley (approved):
Note for posterity: the data_table does not contain the final catalyst SMILES used in the enumeration. The enumeration was performed using the Jupyter notebook, not the standard enumeration script

skearnes:
Thanks for the reviews!

PR #15, opened 2020-10-09
Add a check for the target branch of submissions

skearnes:
Submissions should be merged into new branches rather than `main`. This check won't be marked as "required" since it will flag non-submission PRs as well.


skearnes:
Note that this only affects PRs from forks; our normal PRs to merge submissions into `main` after processing won't be affected.

connorcoley (approved):

PR #23, opened 2020-10-20
Add a badge that counts reactions

skearnes:


connorcoley (approved):
Is it normal to have a bunch of separate jobs when they all have the same kind of dependencies? It seems rather inefficient (even though we don't really need to worry about the cost)

skearnes:
> Is it normal to have a bunch of separate jobs when they all have the same kind of dependencies? It seems rather inefficient (even though we don't really need to worry about the cost)

Yes, I think this is normal. It is inefficient, but I think GitHub is able to cache a lot of this stuff so we can focus on readability.

PR #25, opened 2020-10-20
Use badges/reaction.py from ord-schema

skearnes:
Depends on https://github.com/Open-Reaction-Database/ord-schema/pull/487


connorcoley (approved):

.github/workflows/submission.yml
```diff
@@ -39,7 +39,7 @@ jobs:
     - name: Update reactions badge
       run: |
         cd "${GITHUB_WORKSPACE}"
-        python badges/reactions.py --root=data --output=badges/reactions.svg
+        python ord-schema/badges/reactions.py --root=data --output=badges/reactions.svg
```
connorcoley, line 42: output needs to be `ord-schema/badges/reactions.svg`?

skearnes: No, I'm keeping the output SVG in ord-data; just the script to generate it lives in ord-schema. Does that make sense?

connorcoley: Yes, that seems reasonable

PR #28, opened 2020-10-22
Nanomole flow dataset from 10.1126/science.aap9112

mwleklin:
Data was generated using both web editor and python command line. 
Data from nanomole flow paper from Pfizer.


mwleklin:
Here is the spreadsheet and template. Let me know if you need anything else!

[First_Submission (2).zip](https://github.com/Open-Reaction-Database/ord-data/files/5425754/First_Submission.2.zip)

skearnes:
A few comments from my review of the template and spreadsheet:

* Please use `NANOMOLE` or `MICROMOLE` instead of `MILLIMOLE` for input amounts; see https://github.com/Open-Reaction-Database/ord-schema/blob/2d0b0de338d0a512010d7b2575c819b8ff74606d/ord_schema/proto/reaction.proto#L1002-L1009
* I don't see anything in the paper or SI that would indicate the amount of 4,4′-di-tert-butylbiphenyl that actually made it into each reaction. Would you mind reaching out to the authors for clarification?
* The solvents used for the reactants appear to be somewhat incorrect. For instance, the spreadsheet shows 1,3-diethylbenzene for all of the aryl halide inputs, but 1e-f used water or DMF.

Other notes:

* Please revert the deletion of `ord_data-lfs_temp.txt`

mwleklin:
> A few comments from my review of the template and spreadsheet:
> 
> * Please use `NANOMOLE` or `MICROMOLE` instead of `MILLIMOLE` for input amounts; see https://github.com/Open-Reaction-Database/ord-schema/blob/2d0b0de338d0a512010d7b2575c819b8ff74606d/ord_schema/proto/reaction.proto#L1002-L1009
_Corrected._
> * I don't see anything in the paper or SI that would indicate the amount of 4,4′-di-tert-butylbiphenyl that actually made it into each reaction. Would you mind reaching out to the authors for clarification?
_In the process of reaching out and finding out._
> * The solvents used for the reactants appear to be somewhat incorrect. For instance, the spreadsheet shows 1,3-diethylbenzene for all of the aryl halide inputs, but 1e-f used water or DMF.
_Thanks for noting this as I did mess out some of the solvents. However the paper is also a bit convoluted. as 1a-d are aryl halides and 2d is. I built my excel sheet to reflect this so the order and name were correct for the enumeration. 1e-1g and 2a-2c are the boronates for the reaction._
> 
> Other notes:
> 
> * Please revert the deletion of `ord_data-lfs_temp.txt`

skearnes:
Superseded by #39

PR #29, opened 2020-10-22
Add Santanilla Experiment-2 Dataset pbtxt

michaelmaser:


skearnes:
It looks like you forgot to include the computed yields?

michaelmaser:
> It looks like you forgot to include the computed yields?

Believe it or not, this was actually intentional. After discussions with @connorcoley, I removed the computed yields as I/we don't believe they're truly quantitative (or relative). See discussion on ord-schema #432 [https://github.com/Open-Reaction-Database/ord-schema/issues/432](url). Can discuss more at our meeting tomorrow.

skearnes (approved):

PR #30, opened 2020-10-23
Skip some checks if there are no *.pb files

skearnes:
Fixes a failing test on #29

### DIFF
--- a/.github/workflows/submission.yml
+++ b/.github/workflows/submission.yml
@@ -149,14 +149,16 @@ jobs:
         python ./ord-schema/ord_schema/process_dataset.py \
           --input_file=changed_pbtxt_files.txt \
           --base=upstream/main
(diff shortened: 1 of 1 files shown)


connorcoley:
If the error is caused not by the absence of the `changed_pb_files.txt` file, but by the fact that it's empty, I think it would make more sense to modify the `process_dataset.py` script to handle that case gracefully

skearnes:
> If the error is caused not by the absence of the `changed_pb_files.txt` file, but by the fact that it's empty, I think it would make more sense to modify the `process_dataset.py` script to handle that case gracefully

The issue is actually in the for loop; process_datasets is fine but the file has a single empty character in it that messes up the calls to check_pb.

connorcoley (approved):

PR #32, opened 2020-10-29
Add Santanilla Experiment-2 Dataset pbtxt (#29)

skearnes:
Merging #29 into main


michaelmaser:
Reaction IDs and things look good. Did notice, though, that the pbtxt still has the image URLs instead of bytes Data. Went back and rewrote the Dataset with the updated ipynb and the pbtxt is now almost 5 GB... Is this okay? Or, should we leave it as is?

skearnes:
> Reaction IDs and things look good. Did notice, though, that the pbtxt still has the image URLs instead of bytes Data. Went back and rewrote the Dataset with the updated ipynb and the pbtxt is now almost 5 GB... Is this okay? Or, should we leave it as is?

Let's keep it as is for now; we can update this later if we want.

michaelmaser (approved):

PR #38, opened 2020-11-24
Jli submission photodehalogen

jun-li-2020:
This is the dataset using estimated 5-hr conversion results from 1,152 heteroleptic iridium(III) photocatalysts derived from the original paper.

[Photodehalogen_Merged_dataset_5hrconv.zip](https://github.com/Open-Reaction-Database/ord-data/files/5586688/Photodehalogen_Merged_dataset_5hrconv.zip)


skearnes:
Thanks @jun-li-2020! Would you mind uploading any script or protocol you used to calculate the estimated conversions? It would be good to have that for posterity.

Also I see a couple extra files in the submission that maybe shouldn't be there?
* enumerate_Pd_CN_dataset_final.pbtxt
* enumerate_Suzuki_dataset_final.pbtxt

jun-li-2020:
@skearnes Ok, I will close this PR .  In the next PR I will upload a zip file incl the estimated conversion in an Excel file  and exclude the two CN and Suzuki datasets.

skearnes:
Feel free to just push additional commits to this PR so you don't have to create a new one...

PR #39, opened 2020-11-24
Dataset Commit for Nanomole flow paper

mwleklin:
This is a 5760 enumerated reactions from DOI: 10.1126/science.aap9112
The internal standard amount is still off. 
Made through a combination of interactive web editor and command line enumeration. 
Template reaction and .csv are attached. 
[Submission.zip](https://github.com/Open-Reaction-Database/ord-data/files/5590977/Submission.zip)


skearnes:
Thanks @mwleklin! A few comments:

* Please remove the "NAME resolved by the PubChem API" details from the template; it doesn't apply to SMILES.
* I'm pretty confused by the spreadsheet. I think the SMILES might be right, but e.g. the `Reactant_1_Name` and `Reactant_1_Short_Hand` columns don't always match (see row 4756, for instance). Could you please update the spreadsheet so these values are correct?

(Also, please add new commits to this PR instead of creating a new one.)

mwleklin:
Removed "NAME resolved by the PubChem API"
I double checked the file for accuracy and corrected this name issue.

skearnes:
> Removed "NAME resolved by the PubChem API"
> I double checked the file for accuracy and corrected this name issue.

Thanks; could you please attach the updated spreadsheet?

mwleklin:
Submission files attached. 

[Submission_2.zip](https://github.com/Open-Reaction-Database/ord-data/files/5598493/Submission_2.zip)

skearnes:
Thanks @mwleklin! Just a couple minor comments:

* Please set `addition_order` to 1 for all the inputs; AFAICT they are added concurrently.
* I think the template you uploaded is not the one you used. In particular, the template has MICROMOLE while the submission uses MILLIMOLE. I'd definitely prefer the former, but it looks like the amounts are correct as-is if it'll be a pain to update.

mwleklin:
Made the changes as desired. The addition order was actually the autosampler draw order but I am good with the change as the reaction isn't intended to kick off until it reaches the heated flow loop.

The micromole should be fixed, not sure why it was millimole previously. 

[Submission_3.zip](https://github.com/Open-Reaction-Database/ord-data/files/5623257/Submission_3.zip)

skearnes:
Hey @mwleklin, thanks again and sorry but there were a couple errors introduced in the latest commit:

- The "Ligand" units were changed to MICROMOLE but the value was not adjusted; it still shows as `5e-05`
- (nit) there's an extra space in one of the input keys: `"Boronic Acid "`
- (nit) the "NAME resolved by the PubChem API" details are back; please remove them again

mwleklin:
Made those changes as requested. I seem to figure out new way to break my submission :)

[Submission_4.zip](https://github.com/Open-Reaction-Database/ord-data/files/5625390/Submission_4.zip)

skearnes:
Thanks @mwleklin, really appreciate your contribution here and iterating with me on the reviews!

PR #40, opened 2020-11-24
Jun li 2020 patch 1

jun-li-2020:
Here incl the estimated conversion calculation in an Excel file, reaction template, and the datasets used for enumeration.

[Photodehalogen_Merged_dataset_5hrconv_ver2.zip](https://github.com/Open-Reaction-Database/ord-data/files/5591977/Photodehalogen_Merged_dataset_5hrconv_ver2.zip)


jun-li-2020:
@skearnes Sorry for any inconvenience, I had removed the previous PR before I just saw your quick responses in email.

skearnes:
Note that this is an evolution of https://github.com/Open-Reaction-Database/ord-data/pull/10

Thanks @jun-li-2020! A few more comments:
* I think the [Ir(C^N)2Cl2]2 dimer solution and the N^N ancillary ligand solution should be represented as separate inputs with the same addition order. Right now you have them listed as separate _components_ in the same input.
* It looks like you selected `LITER` instead of `MICROLITER` for the DMSO volume.
* It looks like you selected `MICROLITER` instead of `MILLILITER` as the vessel volume.
* Please include the glass vial details: "9800-840 glass shell vials".

Thanks!

jun-li-2020:
@skearnes Thank you very much!  I have fixed the following issues you have found: 1)DMSO changed to MICROLITER, 2)Vessel volume changed to MILLILITER, 3) Include the glass vial details: "9800-840 glass shell vials".  Now as for the 1st observation on using two separate inputs instead of two components for one single input in the preparation of the photocatalyst, thanks again  for this comment. However, I believe the use of a single heteroleptic iridium catalyst input depicted in the reaction scheme provides a better view for the chemist. It reflects the actual "active" photocatalyst input used in the reaction although it was pre-formed over 150C overnight from two separate components.  If we choose to use two separate inputs, then the telescoped process should have been described with separate reaction IDs and the resulting catalyst from 1st preparative reaction will be used in situ in the 2nd dehalogenation reaction.  This 2nd approach would be a bit complicated.  In my view, the 1st approach is easier to construct and more clear to view schematically without losing the accuracy when custom details are provided for preparing this multi-component input. Therefore, I have the reservation to switch from two-component to the 2-input representation.
[Photodehalogen_Merged_dataset_5hrconv_final.zip](https://github.com/Open-Reaction-Database/ord-data/files/5594069/Photodehalogen_Merged_dataset_5hrconv_final.zip)

skearnes:
Thanks @jun-li-2020; I discussed with @connorcoley and we agree that the catalyst should either be represented (1) as a separate reaction with all inputs separate, or (2) as a single input where the full catalyst is a single component in DMSO.

For this PR, I think option (2) is the most straightforward option. Could you please adjust the catalyst input to have two components: one for the final catalyst (along with details about how it was formed) and one for the DMSO solvent? The catalyst structure should be as close as possible to the actual final form (i.e. no Cl atoms)---dative bonds to the Ir would be ideal, but separate components separated by periods is acceptable as well.

skearnes:
Hi @jun-li-2020, just following up on this. I really appreciate your patience as we're figuring out the review process during alpha testing---let me know if you have any other questions!

jun-li-2020:
@skearnes , Happy Holiday! Thank you for your following up.  As we discussed before in the email regarding the dative bond octahedral image was severely distorted and the folllowing discussion you suggested: "I'm actually less concerned about the drawings; we might be able to improve those over time. If you have the dative bond form working I'd just use that.  In the case of period-delimited components, I'd probably use the neutral forms and include two of the C^N. This most closely matches Figure 1 in the paper (although again I prefer the dative bond form even if it doesn't render nicely). Connor, any additional thoughts?"  @connorcoley Don't you if you have any follow-on comments?  I also thought over the holiday that the way I am currently encode the mol oobject as "Chem.MolFromSmiles('CC(C)(C)C1=CC(=N->4C=C1)C2=N->5C=CC(=C2)C(C)(C)C.C1=CC=N->6C(=C1)C2=C(C=C(C=[C-]->82)F)F.C1=CC=N->7C(=C1)C2=C(C=C(C=[C-]->92)F)F.[Ir]456789', sanitize=False)" would not be appropriately recognized using conventional morgan fingerprints.  Ideally we would like to easily scrape the open rxn database with the smiles for other machine learning studies and the way dative bonds in this complex would not help and people has to clean up the enumerated smiles containing dative bonds to do other computational works. I just felt that there might be more issues by incorporating dative bonds for organometallic catalysts for converting to fingerprints (maybe consult with Greg or Jensen?) or xyz mol file used directly for Gaussian, PSI4 quantum mechanic calculations?  Therefore, I would suggest keeping the individual C^N, N^N smiles in the current enumerate dataset, and possibly we can create a separate enumerated dataset including dative bond in that particular set.  Any thoughts?

skearnes:
Hi @jun-li-2020, I discussed more with @connorcoley and we've decided that until we can settle on a more standard catalyst definition we are just going to be looking to see that all of the relevant information is present. The current dataset, which lists each component of the catalyst separately, is fine.

Unless there are any other comments, I'll plan on merging this later today. Thanks again!

skearnes (approved):

jun-li-2020:
@skearnes @connorcoley Thank you very much for approving this PR!  Best, -Jun

connorcoley:
Thank you for the contribution!

PR #47, opened 2021-01-15
Clean up provenance DOIs (#46)

skearnes:
* Clean up provenance DOIs

* trigger actions

* Update submission.yml


skearnes:
Hold on a sec; I'm confused by the update commit...

skearnes:
I think it merged the changes into a single dataset; let me update that (we need to disable this anyway for sharded datasets).

PR #50, opened 2021-01-18
example submission of liu: 10.15227/orgsyn.095.0080

ipendlet:
Generated using a slightly extended version of the jupyter notebook code from https://github.com/open-reaction-database/ord-schema/tree/main/examples/3_Liu_Copper_OrgSyn.

### DIFF
--- a/2021-01-18_Liu_reactions.pbtxt
+++ b/2021-01-18_Liu_reactions.pbtxt
@@ -1,0 +1,1480 @@
+name: "Copper-Catalyzed Enantioselective Hydroamination of Alkenes"
+description: "Three reactions from: 10.15227/orgsyn.095.0080"
+reactions {
+  inputs {
+    key: "4-dimethyl-aminopyridine"
+    value {
+      components {
+        identifiers {
+          type: SMILES
+          value: "n1ccc(N(C)C)cc1"
+        }
+        identifiers {
+          type: NAME
+          value: "4-dimethyl-aminopyridine"
+        }
+        amount {
+          mass {
+            value: 12.8
+            units: GRAM
+          }
+        }
+        reaction_role: REACTANT
+        preparations {
+          type: NONE
+          details: "used as received"
+        }
+        source {
+          vendor: "Sigma Aldrich"
+        }
+      }
+      addition_order: 2
+    }
+  }
+  inputs {
+    key: "N,N-dibenzylhydroxylamine"
+    value {
+      components {
+        identifiers {
+          type: SMILES
+          value: "C1=CC=C(C=C1)CN(CC2=CC=CC=C2)O"
+        }
+        identifiers {
+          type: NAME
+          value: "N,N-dibenzylhydroxylamine"
+        }
+        amount {
+          mass {
+            value: 21.3
+            units: GRAM
+          }
+        }
+        reaction_role: REACTANT
+        is_limiting: true
+        preparations {
+          type: CUSTOM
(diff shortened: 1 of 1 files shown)


skearnes:
Thanks Ian!

* Adding @connorcoley to review since he wrote the original notebook.
* Adding @brilee for "shadow" review.

2021-01-18_Liu_reactions.pbtxt
```diff
@@ -0,0 +1,1480 @@
...
+        type: PURITY
+        percentage {
+          value: 99.0
+        }
+      }
+      measurements {
+        analysis_key: "isolated_weight"
+        type: AMOUNT
+        amount {
+          mass {
```
connorcoley, line 410: Could potentially add a precision of 0.25 gram here

2021-01-18_Liu_reactions.pbtxt
```diff
@@ -0,0 +1,1480 @@
...
+        details: "dry nitrogen"
+      }
+    }
+    stirring {
+      method {
+        type: STIR_BAR
+      }
+    }
+    conditions_are_dynamic: true
+    details: "reaction started in ice bath before addition of\npivaloyl chloride, then allowed to warm to 23 C, then stirred for \'\nan additional 6 hours at RT (23 degC)"
```
connorcoley, line 193: Looks like there is a stray quotation mark here. The newline symbols `\n` aren't ideal for these long, multi-line strings.

2021-01-18_Liu_reactions.pbtxt
```diff
@@ -0,0 +1,1480 @@
...
+  }
+  provenance {
+    city: "Cambridge, MA"
+    doi: "10.15227/orgsyn.095.0080"
+    publication_url: "http://www.orgsyn.org/demo.aspx?prep=v95p0080"
+    record_created {
+      time {
+        value: "01/18/2021, 10:02:57"
+      }
+      person {
```
connorcoley, line 514: You should update this to include yourself! If you'd like to keep both of us, then you can add a "RecordModified" event

2021-01-18_Liu_reactions.pbtxt
```diff
@@ -0,0 +1,1480 @@
...
+        name: "Connor W. Coley"
```
connorcoley, line 993: Same comment as above re: provenance

2021-01-18_Liu_reactions.pbtxt
```diff
@@ -0,0 +1,1480 @@
...
```
connorcoley, line 1473: Same comment

connorcoley (changes requested):
Thanks for filling out this example! I've made a couple _very minor_ comments in-line.

I have one larger suggestion, which will actually be very helpful to test an aspect of the schema that isn't often used. For the second and third reactions, we use the pivaloylhyodroxylamine product from the first reaction. This component appropriately has type "SYNTHESIZED" as its preparation in both cases. However, we can be even more explicit and add a reference to the specific reaction that produced it. To do this, you should be able to give the first reaction a placeholder string in its ID field and refer to that in the SYNTHESIZED preparation. Upon submission, when validating the whole dataset, the script should automatically update that cross-reference with the newly-assigned reaction ID. Would you be willing to give that a shot?

skearnes:
Hi @ipendlet, it's been a while since the last activity on this PR. Do you have any questions we can answer?

PR #51, opened 2021-02-04
Uploading dataset pbtxt for Shields Doyle Paper

Jesus00-1:
Pbtxt files were created using the interactive web editor. Reactions are from the paper https://doi.org/10.1038/s41586-021-03213-y. 
Template reactions pbtxt and csv files are attached.

Note 0% Yields did not show up in the schema when checking for consistency.

This is an enumerated dataset of 1728 reactions from imidazole arylation for figure S42.

[Shields_arylation_figS42.zip](https://github.com/open-reaction-database/ord-data/files/5927349/Shields_arylation_figS42.zip)

This is an enumerated dataset of 256 reactions from imidazole arylation for figure S45 (Extended ligand set for testing model out of sample prediction performance)

[experiment_index_extended_dataset.zip](https://github.com/open-reaction-database/ord-data/files/5927409/experiment_index_extended_dataset.zip)

PR #57, opened 2021-02-08
Shields_Arylation

Jesus00-1:
Shields Arylation S42 and S45. Paper: https://doi.org/10.1038/s41586-021-03213-y

Best,
Jesus


Jesus00-1:
I merged into the main branch, but the check is failing. Which branch in the repo should I merge to?

Best,
Jesus

Jesus00-1:
uploading the zip files.

[experiment_index_extended.zip](https://github.com/open-reaction-database/ord-data/files/5947557/experiment_index_extended.zip)
[experiment_index_S42.zip](https://github.com/open-reaction-database/ord-data/files/5947558/experiment_index_S42.zip)

skearnes:
> I merged into the main branch, but the check is failing. Which branch in the repo should I merge to?
> 
> Best,
> Jesus

That check is ok to fail; I'll change the target branch when we're ready to merge. The Submission/process_submission check is the important one.

A couple things:
- It looks like you committed the template reaction pbtxt; please remove that.
- The second pbtxt file is much smaller than the one from #51; just checking that you're only intending to submit the 60k-line file?

michaelmaser (changes requested):
Looks good for the most part. Couple things:

1. The vessel type given is `WELL_PLATE`, which is true, though the reactions are actually set up in `VIAL`s of material `GLASS` in a well plate, not in the plate wells directly. @skearnes @connorcoley, don't know that we've discussed this yet. How should this be captured? Also, the pressure control is listed as `SEALED`, but no `VesselAttachment` is listed. I assume the vials are capped, which can be captured by adding a `CAP` `VesselAttachment`; please add this.
2. There is only one workup step (`ALIQUOT`) included, which contains string details describing several more steps. The full workup is described as: 

> The plate was removed from the glovebox, opened, and diluted to a 900 uL total volume with N,N-dimethylacetamide. The plate was stirred for 5 min and a 75 uL sample was taken and filtered into an HPLC analysis plate. The filter was rinsed with 400 uL acetonitrile/water (4:1) solution and analyzed by UHPLCMS. 

This should be captured by the full workup step sequence: `ADDITION` -> `STIRRING` -> `ALIQUOT` -> `FILTRATION` -> `ADDITION`.

3. The `INTERNAL_STANDARD` is listed in a separate `reaction.outcome`, though it is analyzed in the same `analyses` key as the desired product. This can be captured by adding the internal standard to the set of products in the first (and only) outcome. Its peak area `measurement` can also be added and linked to the same `UHPLCMS` analysis that the product is measured with. These values are not included in the dataset, only the ultimate processed reaction yields. I don't know that this is necessarily a problem (@skearnes @connorcoley?), but do you have the raw peak areas for the internal standard?

michaelmaser (approved):
Looks great. Thanks for working through this, Jesus!

PR #60, opened 2021-02-11
Add USPTO-50K dataset from https://doi.org/10.1021/ci5006614

skearnes:
I had to create dummy inputs/outputs to get the validations to pass. I assume we're more concerned about easy access than avoiding duplicates for these cases where we only include the reaction SMILES.

I've attached the notebook I used to replicate the train/test split.

[uspto-50k.zip](https://github.com/open-reaction-database/ord-data/files/5968962/uspto-50k.zip)


skearnes:
Oops, the [data](Development_of_a_Novel_Fingerprint_for_Chemical_Reactions_and_Its_Application_to_Large_Scale_Reaction_Classification_and_Similarity) has a CC-BY-NC license; we'll need to ask them for permission to include.

PR #61, opened 2021-02-12
Add USPTO-480K dataset from https://doi.org/10.1039/C8SC04228D

skearnes:
Here's the notebook I used to create these: [uspto-480k.zip](https://github.com/open-reaction-database/ord-data/files/5996896/uspto-480k.zip). Only the reaction SMILES are included; pending https://github.com/open-reaction-database/ord-schema/pull/559 for the validations to pass.

The data in https://github.com/connorcoley/rexgen_direct is licensed under GPL-3.0, and @connorcoley has consented to re-license for inclusion in the ORD.


skearnes:
Updated to use only reaction SMILES and not add in made-up stuff to make the validations happy; PTAL.

connorcoley:
The information after the reaction SMILES is an annotation of bond changes during the reaction, not actually information related to the reaction following the extended SMILES format. Everything after the space should just be discarded and the type changed to plain old `REACTION_SMILES` (also the description updated to reflect that these aren't REACTION_CXSMILES)

skearnes:
> The information after the reaction SMILES is an annotation of bond changes during the reaction, not actually information related to the reaction following the extended SMILES format. Everything after the space should just be discarded and the type changed to plain old `REACTION_SMILES` (also the description updated to reflect that these aren't REACTION_CXSMILES)

Thanks; fixed.

connorcoley (approved):

PR #79, opened 2021-03-30
Add dataset from Novartis

skearnes:
From [Fabio Lima](https://www.linkedin.com/in/lima-fabio/) at Novartis.


limafabio2:
Thank you @skearnes for doing the pull request, I hereby confirm that I agree with sharing this datasets under the Licence and Terms of the ORD.

connorcoley:
This submission looks great, and it's exciting to get unpublished work in here! Some **optional** suggestions for @limafabio2 to consider:
- Because the analysis _names_ are just for making cross-references, I would suggest that "LCMS-CAD" be included in the _details_ field of the analysis, preferably with the instrument manufacturer as well so we know the exact detector used
- Having a yield value reported as a percentage that (a) doesn't use an internal standard, (b) isn't normalized, and (c) doesn't use an authentic standard makes it a little tricky to know how much it can be trusted. For this case, is the yield value actually normalized to 0-100% based on the expected mass and some universal CAD calibration curve? If yes, I would consider this "normalized"
- If possible, could you mention what automated platform was used (under Setup)?

If it is easier for you, we can make those tweaks if you provide the original template pbtxt, enumeration spreadsheet, and these extra details (here or by email)

skearnes:
Updated with Connor's additions via #80

connorcoley (approved):

PR #84, opened 2021-04-28
Coupling of α-carboxyl sp3-carbons with aryl halides

connorcoley:
Substrate scopes (Figure 3 and 4A) from https://science.sciencemag.org/content/345/6195/437

Files attached. Lots of manually copying analytical/outcome data from the SI, unfortunately, in `photoredox_data_analytical.csv`
[Archive.zip](https://github.com/open-reaction-database/ord-data/files/6395066/Archive.zip)


brilee:
Comments:
- Ir should be in the [3+] oxidation state. 
- I take it that the C-H activation example wasn't digitized, just the arene and amide scopes?
- For the 3x ether extractions, I noticed the workup type is "ADDITION"; shouldn't it be EXTRACTION?
- I checked all product structures; LGTM
- I double-checked to see if there was any wavelength information; seems like that is kind of important for a paper on photocatalysis... didn't see any, though.

brilee:
Agreed with [Ir+]. I was initially confused since the pyridines are a charge-neutral ligand, but there's actually an anionic aryl ligand with a direct C-Ir bond, so that makes the overall charge of the iridium center +1.

connorcoley:
Fixed the Ir, PTAL and approve if it's okay @brilee

brilee (approved):
LGTM

PR #86, opened 2021-05-24
Add hein-dataset.pbtxt

skearnes:
Multi-step reactions from https://pubs.rsc.org/en/content/articlelanding/2019/re/c9re00086k. Measurements of the reaction composition were taken over time and are each encoded as a separate outcome. I used two identical analyses to distinguish between online and offline measurements.

Supporting files: [hein-submission.zip](https://github.com/open-reaction-database/ord-data/files/6549700/hein-submission.zip) (updated)


connorcoley (changes requested):
- [ ] - Synthesis of intermediate 2 looks great; it’s always fun to see when the full level of detail can be captured by the schema. I infer that the yield was measured by isolated weight (although the weight isn’t reported) — perhaps it is worth adding a WEIGHT analysis?
- [x] - Great to see the addition_duration fields get used
- [x] - For 3-(E), I think we need to set conditions_are_dynamic to True because of the unconventional temperature schedule (20 min at -78 then warmed to 25 and left for 2 hours)
- [x] - For 3-(E), you could note that the reaction “forms_precipitate” and “is_exothermic” (inferred from the slow addition and comments about maintaining an internal temperature below -60 while in a -78 bath)
- [x] - For 3-(E), I don’t know where you got the 16 hour reaction time from. I might consider the reaction time 20 minutes at -78 with a 2 hour WAIT workup at 25 C?
- [ ] - For the saturated NaCl solution, I assumed you just looked up the solubility? I’ve done that before, too. This is one area where perhaps the requirement to define amounts works against us; this isn’t an uncommon scenario. I don’t have a recommendation for this PR, but something for us to consider in general
- [ ] - The SI does not seem to say how they isolated 3-(E) from 3-(Z) and/or how the yields were individually quantified. This would be a nice detail to add if we knew.
- [x] - Pd catalysts seem fine
- [ ] - I would have had a preference for reporting amounts as masses rather than moles, but I won’t make you change all of them. For the future, it’s much more natural. e.g., for Procedure 1, 13 mol versus 2.17 mg (presumably the mass was how the experiment was performed)
- [x] - I didn’t see the automated sampling procedure described in detail (using the dilution plate, etc.) but that’s probably okay. 
- [ ] - I was contemplating whether we should make “Concentration” a structured measurement, but I think the answer is probably not. It’s not like concentration is what is measured directly; it is just a derived property. It’s also very rarely what is measured in these analyses, except for kinetic profiling studies. We might want to discuss this further. 
- [x] - SMILES all look good to me
- [x] - It took me a while, but I finally understand what you’re doing with data masks to split the spreadsheets into multiple experiments. For future reference, I wouldn’t be opposed to doing a little bit of Excel clean-up so each reaction is on its own sheet (or in its own file). I realize you were trying to work with the raw data as provided, and what you did seems to be the best solution for that!

skearnes:
Thanks Connor for the careful review! I made many of the requested changes and added comments below for the others.

> * [ ]  - Synthesis of intermediate 2 looks great; it’s always fun to see when the full level of detail can be captured by the schema. I infer that the yield was measured by isolated weight (although the weight isn’t reported) — perhaps it is worth adding a WEIGHT analysis?

How much do we want to add inferred steps that are not explicit in the methods? I'm happy to do this but I want to make sure there's value in having an empty WEIGHT section.

> * [ ]  - For the saturated NaCl solution, I assumed you just looked up the solubility? I’ve done that before, too. This is one area where perhaps the requirement to define amounts works against us; this isn’t an uncommon scenario. I don’t have a recommendation for this PR, but something for us to consider in general

Yes, I just looked it up; I agree it's not ideal and it might be better to do something like describe it in the details, but then we run into the validations again with missing amounts...

> * [ ]  - The SI does not seem to say how they isolated 3-(E) from 3-(Z) and/or how the yields were individually quantified. This would be a nice detail to add if we knew.

Agreed; I don't see anything there. There might be some info in https://pubmed.ncbi.nlm.nih.gov/26743694/ (cited in footnote 1 of the SI) but that's behind another paywall for me and not in their SI.

> * [ ]  - I would have had a preference for reporting amounts as masses rather than moles, but I won’t make you change all of them. For the future, it’s much more natural. e.g., for Procedure 1, 13 mol versus 2.17 mg (presumably the mass was how the experiment was performed)

Thanks; will keep in mind for the future.

> * [ ]  - I was contemplating whether we should make “Concentration” a structured measurement, but I think the answer is probably not. It’s not like concentration is what is measured directly; it is just a derived property. It’s also very rarely what is measured in these analyses, except for kinetic profiling studies. We might want to discuss this further.

Agreed.

skearnes:
Here's the updated notebook: [hein-submission.zip](https://github.com/open-reaction-database/ord-data/files/6750301/hein-submission.zip)

connorcoley (approved):
> How much do we want to add inferred steps that are not explicit in the methods? I'm happy to do this but I want to make sure there's value in having an empty WEIGHT section.

If they purify the product before reporting the yield, treating it as an isolated yield based on mass is a safe assumption in my opinion. I guess the value of having an empty WEIGHT analysis would be that we know it is an isolated yield, since otherwise we don't have _any_ analysis_key that tells us if we can trust the measurement. Since they didn't report the mass and theoretically could have gotten the yield through some other method, we can leave the current measurement as-is. 

Thanks for all of the other changes and replies!

PR #95, opened 2021-06-07
OrgSyn example from ord_schema

connorcoley:
Exactly uses the example notebook from ord_schema. Notebook and pbtxt attached

[Archive.zip](https://github.com/open-reaction-database/ord-data/files/6612310/Archive.zip)


michaelmaser (changes requested):
Mostly minor comments, besides workup for reaction 2 & 3.

Reaction 1:

- [ ] DMAP has `reaction_role` = `REACTANT` - Think I would consider it a `REAGENT` since it's not contributing atoms to the product.
- [ ] You have 3 `ADDITION`s of 50 mL DCM in the workup, then an `EXTRACTION` with details of "3 x 50 mL". It seems like the procedure has 1 addition (to wash the reaction flask on transferring to the sep funnel) then 2 extractions. Is this how we handle extractions (`ADDITION`, then `EXTRACTION`)? Or, should they just be `EXTRACTION`s?
- [ ] Could the `ADDITION` step between the concentration and chromatography be considered a `DISSOLUTION`?
- [ ] I think I would consider the purification `FLASH_CHROMATOGRAPHY`. They load onto a column and collect fractions as usual, just that the stationary phase is alumina (our note in the proto says to specify this). Seems like we added `OTHER_CHROMATOGRAPHY` for things like prep TLC, prep HPLC and things.
- [ ] The analytical data is great
- [ ] Want to add experimenter name to provenance?

Reaction 2:

- [ ] Thinking reaction_role for Cu(OAc)2, DTBM-SEGPHOS, and PPh3 could probably be `CATALYST`
- [ ] N,N-dibenzyl-O-pivaloylhydroxylamine has preparation `SYNTHESIZED`, but doesn't reference reaction 1, want to add this?
- [ ] Want to specify this reactant `is_limiting=False`? Have it noted for the styrene and could be derived from amounts, just noting.
- [ ] dimethoxy(methyl)silane has addition order 5, but was added after THF (addition order 6)
- [ ] setup and observation details are nice
- [ ] The workup seems to be missing?
- [ ] Funny consequence of using Python floats for HRMS data, making note here to remind myself to create issue later. Precision/number of sig figs is way beyond instrument precision, and could change the resulting delta between expected and measured. In the worst case scenario, this change would bring a measurement outside the acceptable range for publication. Not crucial right now, but worth noting for auto SI generation down the line. Possible solutions are to handle as strings or round to 4 decimals. Found same thing for isolated_weight in analysis.
- [ ] Analytical data is really nice

Reaction 3:

- [ ] Basically all the same comments from step 2
- [ ] Yield in the paper is 87%, vs 86% in the dataset, perhaps copied from reaction 2

connorcoley:
> Mostly minor comments, besides workup for reaction 2 & 3.
> 
> Reaction 1:
> 
> * [X]  DMAP has `reaction_role` = `REACTANT` - Think I would consider it a `REAGENT` since it's not contributing atoms to the product.
> * [ ]  You have 3 `ADDITION`s of 50 mL DCM in the workup, then an `EXTRACTION` with details of "3 x 50 mL". It seems like the procedure has 1 addition (to wash the reaction flask on transferring to the sep funnel) then 2 extractions. Is this how we handle extractions (`ADDITION`, then `EXTRACTION`)? Or, should they just be `EXTRACTION`s?

I think this has always been vague. To me, doing 3 additions then an extraction makes the most sense because otherwise the process isn't linear. When I think about the linear workup steps, keep_phase means _that_ material becomes what we're tracking throughout the rest of the workup. This is tricky since our workups are linear sequences but this is not a sequential operation.

> * [X]  Could the `ADDITION` step between the concentration and chromatography be considered a `DISSOLUTION`?
> * [X]  I think I would consider the purification `FLASH_CHROMATOGRAPHY`. They load onto a column and collect fractions as usual, just that the stationary phase is alumina (our note in the proto says to specify this). Seems like we added `OTHER_CHROMATOGRAPHY` for things like prep TLC, prep HPLC and things.
> * [X]  The analytical data is great
> * [X]  Want to add experimenter name to provenance?

How do we know if Richard or Steve ran the reactions? Just kidding, I'll add Richard as the experimenter for all three
> 
> Reaction 2:
> 
> * [X]  Thinking reaction_role for Cu(OAc)2, DTBM-SEGPHOS, and PPh3 could probably be `CATALYST`

That's fair. Changed

> * [X]  N,N-dibenzyl-O-pivaloylhydroxylamine has preparation `SYNTHESIZED`, but doesn't reference reaction 1, want to add this?
> * [X]  Want to specify this reactant `is_limiting=False`? Have it noted for the styrene and could be derived from amounts, just noting.
> * [X]  dimethoxy(methyl)silane has addition order 5, but was added after THF (addition order 6)

Thanks, not sure why I flipped those originally

> * [X]  setup and observation details are nice
> * [X]  The workup seems to be missing?

I swear I had added those to the Jupyter notebook months ago, but it doesn't look like it. The new version should have it.

> * [X]  Funny consequence of using Python floats for HRMS data, making note here to remind myself to create issue later. Precision/number of sig figs is way beyond instrument precision, and could change the resulting delta between expected and measured. In the worst case scenario, this change would bring a measurement outside the acceptable range for publication. Not crucial right now, but worth noting for auto SI generation down the line. Possible solutions are to handle as strings or round to 4 decimals. Found same thing for isolated_weight in analysis.
> * [X]  Analytical data is really nice
> 
> Reaction 3:
> 
> * [X]  Basically all the same comments from step 2
> * [X]  Yield in the paper is 87%, vs 86% in the dataset, perhaps copied from reaction 2


[Archive.zip](https://github.com/open-reaction-database/ord-data/files/6671177/Archive.zip)

michaelmaser (approved):
LGTM. Thanks for the changes

PR #98, opened 2021-06-15
Electroreductive coupling of Alkenyl and Benzyl Halides

brilee:
[delano.zip](https://github.com/open-reaction-database/ord-data/files/6658941/delano.zip)


connorcoley:
- [ ] It looks like you omitted (S,S)-3k and (R,S)-3k from the alkenyl bromide scope. Was this intentional? It would be nice to have that reaction for completeness, or at least comment in the dataset description that it was omitted. To the description, could you also clarify that this data represents Figure 2 and Figure 3?
- [X] Reactants and ligand in the ChemDraw file all look good
- [X] Product stereochemistry looks correct form SMARTS string
- [ ] "DeLano" as a reaction identifier might not be appropriate. If you'd like a reaction name, perhaps a general type description would be better, like "nickel-catalyzed reductive cross-coupling" or "enantioselective nickel-catalyzed reductive cross-coupling of alkenyl bromides and benzyl chlorides" (even though this is rather long)
- [ ] The 3 mL of DMA could make use of the SYRINGE addition device
- [ ] Likewise, the benzyl chloride addition could make use of the SYRINGE addition device and ALL_AT_ONCE speed. The addition_order for this input should be 5, not 3
- [ ] Does the procedure specify how the reaction was cooled to 0 C or are you assuming it was an ice bath?
- [X] The electrochemistry conditions look great! I'm glad we're getting to use that part of the schema
- [ ] The quench with 1N aqueous HCl described in the SI (Procedure 1) does not seem to be listed in the workup steps
- [ ] Right now, we can't describe concentrations of solutions without a volume amount. This seems to be a little more frequent in workup procedures than I anticipated. How should we think about addressing this (CC @michaelmaser )?
- [ ] The detailed analytical data for each product is omitted. This is probably okay since we have the summary metrics of reaction performance (yield and ee), but just a note for posterity. It would be rather tedious to include all of this information for these _retrospective_ data entries using already-published dat, but we should try to include them as much as possible in the future. 
- [ ] It could be worth at least noting that 1H NMR, 13C NMR, FTIR, and HRMS were all taken for these products even if you don't copy all of the actual data or peaks over to the entry. (also CC @michaelmaser for opinion)
- [X] Yields and EEs look good and are recorded properly
- [ ] If you would like, you could add details about the SFC instrument and method to that analysis


We should get your ord_schema PR reviewed soon since this submission uses the new macros.

connorcoley (changes requested):
See above

brilee:
re: cool to 0 with ice bath. The SI does not explicitly say "ice bath" but my understanding is that certain magic numbers mean certain cooling conditions (e.g. a -78C would indicate dry ice/acetone). 0 means ice bath to me, in a benchtop setting.

re: detailed analytical data. Would you like me to add the analytical data or skip?

brilee:
[delano.zip](https://github.com/open-reaction-database/ord-data/files/6744374/delano.zip)
Latest supporting info for this PR

brilee:
@skearnes the check_file_types test is failing but I can't figure out why

connorcoley:
Thanks for adding the diastereoselective compounds!

To the description, could you also clarify that this data represents Figure 2 and Figure 3? There are also more than 24 members in this library, so the description isn't quite accurate.

> re: cool to 0 with ice bath. The SI does not explicitly say "ice bath" but my understanding is that certain magic numbers mean certain cooling conditions (e.g. a -78C would indicate dry ice/acetone). 0 means ice bath to me, in a benchtop setting.

Since they don't explicitly say it though, we should probably just record that the temperature was 0 C and not make that logical connection. They could have used a dry cooling block in theory.

> re: detailed analytical data. Would you like me to add the analytical data or skip?

I do think it would be nice to add an Analysis for these different types as an indication that the information might be available in the original paper. You don't have to provide the actual data or details, but noting the existence of the analyses would be worthwhile. Because the SFC was used for EE calculation for almost all of these species, copying/pasting its description into the details would be good, too.

> @skearnes the check_file_types test is failing but I can't figure out why

This is likely because the comparison is made to the main branch of ord-data. This branch (#98) is now out of date, so the reaction badge svg is identified as a changed file. If you update your branch with the most recent version of ord-data/main, this should be resolved

brilee:
Progress so far:
- Description changed to 27 members, with figs 2, 3.
- Ice bath removed
- Placeholder entries for 1H, 13C are added, with descriptions but no data.
- Optical rotation data is available, but I have not yet added an analysis, as the AnalysisType enum doesn't have an entry for it.

brilee:
[delano.zip](https://github.com/open-reaction-database/ord-data/files/6858935/delano.zip)
Latest supporting info for the electrochem dataset

brilee:
[delano.zip](https://github.com/open-reaction-database/ord-data/files/6859007/delano.zip)
with the negative float validation fix

connorcoley:
Last request, I promise -- it seems like the dataset description wasn't actually updated in the dataset or notebook. Do you mind double-checking that? Then we're good to go!

brilee:
My vscode has been flaky about saving ipynb files for some bizarre reason :( That edit must have gotten dropped at some point. Thanks for catching it! I believe it should be fixed now.

connorcoley (approved):

brilee:
Any special steps required for me to merge, or should I just hit the "Merge" button?

connorcoley:
Just merge!

Then, create a fresh PR to merge the branch into main. You may need to close/open that PR once or twice to get the actions to run properly (parses dataset, assigns reaction/dataset IDs, validates, counts, etc.). When all the checks pass, add me as a reviewer and I'll approve it for the final merge

PR #103, opened 2021-06-26
Flow dataset from https://pubs.acs.org/doi/full/10.1021/co400012m

connorcoley:
39 reaction library listed in Table 2. Details are rather sparse, but this is a nice library generated in flow through a single-step reaction.

Template reaction made using the online editor & manually changing yield entries to be placeholders. Excel sheet with all reactants (mols drawn in Chemdraw) and yields. Short script used to enumerate product SMILES. 

Note that there is more detail in the paper on product outcomes (states, colors, analytical data for isolated products) than I've copied. The two performance metrics I focused on were those in Table 2: crude NMR yield and isolated yield after workup. We're missing details about the pump type, reactor tube diameter, and how it was warmed. But this at least uses the addition flowrate yield and flow condition details.

[Archive.zip](https://github.com/open-reaction-database/ord-data/files/6720705/Archive.zip)

### DIFF
--- a/dataset.pbtxt
+++ b/dataset.pbtxt
@@ -1,0 +1,14082 @@
+name: "39 compound library from \"Building a Sulfonamide Library by Eco-Friendly Flow Synthesis\""
+description: "Library generated in DOI\302\24010.1021/co400012m (Table 2)"
+reactions {
+  inputs {
+    key: "amine and nahco3 in h2o/peg"
+    value {
+      components {
+        identifiers {
+          type: NAME
+          value: "NaHCO3"
+        }
+        identifiers {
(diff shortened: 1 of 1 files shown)


michaelmaser (approved):
LGTM. Agree sparse in detail but procedure is pretty simple; suppose that's sort of the point.

Loaded the reactions_out.csv and checked product structures, looks good. Yield data all matches and reaction renderings in the editor match for a random handful of reactions. Agree think only things to add would be color details and NMR peaks, but not super relevant for this dataset and we've demonstrated them in others. Don't think I have anything to change.

Thanks!

PR #111, opened 2021-07-08
Update DOI list

skearnes:


connorcoley (approved):
We've got a duplicated entry here for Kutchukian et al. (two ORD datasets, it seems) -- not sure how best to resolve

skearnes:
> We've got a duplicated entry here for Kutchukian et al. (two ORD datasets, it seems) -- not sure how best to resolve

Handled in #112

connorcoley (approved):

PR #119, opened 2021-10-13
uploading Dataset pbtxt for AstraZeneca dataset

bznan:
This is an enumerated dataset of 750 reactions generated from AstraZeneca ELN (10.26434/chemrxiv.14589498.v2) created using the interactive web editor. The pbtxt file is attached.


bznan:
Hi, sorry I am new to this. I found there is one reactant smiles missing in the validation report. I have updated the data file, is it okay to reopen it now?

PR #120, opened 2021-10-14
uploading pbtxt file generated from AstraZeneca ELN

bznan:
A pbtxt file including 750 reactions generated from AstraZeneca by using web editor. The updated file solved one missing SMILES string error.

### DIFF
--- a/AZ_dataset.pbtxt
+++ b/AZ_dataset.pbtxt
@@ -111,7 +111,7 @@ reactions {
     }
   }
   notes {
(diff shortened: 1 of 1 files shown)


skearnes:
Thanks @bznan, the tests are passing now!

connorcoley:
Thanks!

If the vessel, setup, stirring, etc. are all currently stored as unstructured text and it would be a manual procedure to go through every single one to structure it, I might suggest just trying to place it in the `procedure_details` field so the information is not lost. Would that be possible?

It looks like the raw format is the UDM! @skearnes and I were recently discussing the prospects of writing an automated translator, so I would be interested in learning more about your experience. Ideally, we could have a lossless 1:1 mapping *if the data we are structuring is the same*

bznan:
@connorcoley  I have updated the reaction type, created date, and put all the text in the procedure_details. I would say there is still a lot of information that could be generated including Simple ID Reaction ID Sample REF for each reaction component. By far it's kind of hard for me to use a template and a spreadsheet to do that I think the mapping you mentioned could be a good solution. The data structure is kind of the same, except for the unstructured text note. And also there is a lot of data curation work needed for this dataset, as you can see in the last part of the dataset the reagents and reactants were not clearly classified, and also the yield information is kind of messy.

bznan:
@connorcoley I first tried using dollar signs but failed, probably because the data type of the input. I found it worked when I directly copy and paste the note in the web editor, so I put in the 'Note' and replace them all by using python script.

connorcoley:
It looks like there is (only?) one ValidationError stopping us from merging this in, which is the string used for the created date. You've got strings like `"Timestamp('2008-07-01 00:00:00')"` which can't be parsed. If you simply convert these to strings like `"2008-07-01"` we should be all set!

bznan:
@connorcoley Hi, sorry to inform you. I found there is a problem related to catalyst amount. Currently, the catalyst amount and the metal amount are the same, but actually, they are not. I have corrected the problems, but how can I update that on Github? Should I open a new request? So sorry for that.

PR #128, opened 2022-02-22
Cleaning formating issue in the freetext of AZdata

bznan:
Formatting issues previously shown in the free text portion of the AZ dataset due to conversion from html to plain text have been removed. All the reaction ID of 750 data points and dataset ID remains the same.

PR #129, opened 2022-03-02
Cleaning formating issue in the freetext of AZdata (#128)

connorcoley:
The formating issue due to translation from html to plain text ,which was shown in the freetext of previous AZ dataset has been cleaned. All the reaction ID of 750 data points and dataset ID remains the same.


skearnes (approved):

skearnes (changes requested):
check w/ authors about new typos

PR #131, opened 2022-03-04
Bzn (#130)

connorcoley:
* Cleaning formating issue in the freetext of AZdata

The formating issue due to translation from html to plain text ,which was shown in the freetext of previous AZ dataset has been cleaned. All the reaction ID of 750 data points and dataset ID remains the same.

* Delete ord_dataset-00005539a1e04c809a9a78647bea649c.pb.gz

* Changing plaintext issue of original AZ dataset

All the reaction ID of 750 data points and dataset ID remains the same. The formating issue due to translation from html to plain text ,which was shown in the freetext of previous AZ dataset has been cleaned.


connorcoley:
@skearnes a spot check looked good to me

skearnes (approved):

PR #145, opened 2022-12-02
Split reaction SMILES for USPTO subset (#124)

skearnes:
* bump ord-schema version

* split reaction SMILES for USPTO subset

* add amounts

* bump ord-schema

* revert ci/cd

Co-authored-by: EC2 Default User <ec2-user@ip-172-30-4-170.ec2.internal>


skearnes:
I think I'm going to reshard these first...

ipendlet (approved):
Viewed all the files. The structure appears to be correct for showing up on searches, but hard to tell without approving and working through the interface.

PR #165, opened 2023-06-08
Add Baumgartner et al. Suzuki reaction dataset

marcosfelt:
Added a dataset for the reaction optimization from this paper: https://pubs.rsc.org/en/content/articlelanding/2018/RE/C8RE00032H

Code used to generate dataset is here: https://github.com/sustainable-processes/multitask/blob/69606a102b981bf2c207efd72741503fc4a349e3/multitask/etl/etl_baumgartner_suzuki.py


skearnes:
These reactions seem to be missing provenance---maybe you need to add a new commit after updating the code?

skearnes:
@marcosfelt I'm closing this due to inactivity.

PR #167, opened 2023-06-08
Add Reizman Suzuki dataset

marcosfelt:
Add Suzuki coupling reaction dataset from Reizman et al.: http://doi.org/10.1039/C6RE00153J

Code used to generate dataset is here: https://github.com/sustainable-processes/multitask/blob/69606a102b981bf2c207efd72741503fc4a349e3/multitask/etl/etl_reizman_suzuki.py


skearnes:
Hi @marcosfelt, thanks for the submission! Just a couple of comments:
* It would be good to mark the product measurements as using an internal standard (since you included one in the inputs); see https://github.com/open-reaction-database/ord-schema/blob/main/proto/reaction.proto#L953
* I think you forgot to call `add_provenance()` in your script?

PR #175, opened 2023-08-21
add amidation dataset

beef-broccoli:
Dataset of a amide coupling reaction with various activators, bases, solvents. 

Associated with [publication](https://chemrxiv.org/engage/chemrxiv/article-details/64c7e2e1658ec5f7e5808425), Fig. 5 and related texts in the manuscript. Experimental details can be found in SI Section 10.

Reaction template, result spreadsheet and dataset pbtxt files are also included. 

[submission 08-21-23.zip](https://github.com/open-reaction-database/ord-data/files/12403326/submission.08-21-23.zip)


connorcoley:
Thanks!!

It is not essential to make this change, but the pre-stirring for 30 minutes before addition of amine could be represented by the `addition_time` field for the amine input. The `addition_order` is the typical thing to specify, as you've done, but there's even more detail that could be included with `addition_time`. That would let us set `conditions_are_dynamic` back to False.

The SI specifies "overnight", which in the template is 16 hours while Figure 5 shows 24 hours -- do you know which it should be? Everything else looks perfect to me in terms of procedure. I spot checked a few structures only.

beef-broccoli:
Thanks for the feedback! Addition time is changed for the amine, and reaction time is corrected to 24 hours (i guess our procedure also was not clear on reaction time). 

One potential issue i spotted is for some of the products, maybe because of the conversion from inchi to smiles, the amide structure sometimes show up weird, with the double bond put on the C-N bond. Only some of the structures have this issue. Is this something we want to fix? I can try to clean it up

connorcoley:
Oh that's a good call. Roundtripping an iminol through RDKit won't automatically convert it to the main amide tautomer. It might be nice to do that conversion now, when we know it's "safe", as opposed to expecting downstream users to. It will also help with search/etc.

You should be able to do this with...
```
import rdkit.Chem as Chem
from rdkit.Chem.MolStandardize import rdMolStandardize
new_mol = rdMolStandardize.CanonicalTautomer(Chem.MolFromSmiles(smi))
new_smi = Chem.MolToSmiles(new_mol)
```

beef-broccoli:
All the amide structures are fixed! Thanks for the help 

[submission 08-31-23.zip](https://github.com/open-reaction-database/ord-data/files/12488819/submission.08-31-23.zip)

PR #183, opened 2024-07-03
dataset submission for item 47463309 from DOI 10.1038/s41467-023-42446-5

qai222:
1430 reactions from 10.1038/s41467-023-42446-5
(only the HTE reactions are included)


bdeadman:
It looks like process_submission check is failing because the submission workflow is using ord-schema 0.3.65 which pre-dates the addition of PIPETTE enums.

```
ValueError: error parsing dataset.pbtxt: 44:15 : '        type: PIPETTE': Enum type "ord.ReactionInput.AdditionDevice.AdditionDeviceType" has no value named PIPETTE.
Error: Process completed with exit code 1.

```

Possibly solve by updating env values in the [ord-data workflows ](https://github.com/open-reaction-database/ord-data/tree/main/.github/workflows).

The check_target_branch check could also be due to the ord-schema version difference?

check_target_branch
```
Run echo "Target branch: ${GITHUB_BASE_REF}"
  echo "Target branch: ${GITHUB_BASE_REF}"
  test "${GITHUB_BASE_REF}" != "main"
  shell: /usr/bin/bash -e {0}
  env:
    ORD_SCHEMA_TAG: v0.[3](https://github.com/open-reaction-database/ord-data/actions/runs/9771576537/job/26989249049?pr=183#step:2:3).65
Target branch: main
Error: Process completed with exit code 1.
```

skearnes:
I'll fix the version today.

skearnes:
See #184

skearnes:
@qai222 The DOI test failures are a bug; I'll send a PR to ord-schema

skearnes:
@bdeadman can you take a look at this dataset for QC?

bdeadman:
Dataset validates and a spot check of the pbtxt file showed no obvious problems. Merging into main.

skearnes:
Thanks. So what happens next is that we create a PR to main from the branch
we merged into. We'll need to open and close the PR a couple of times since
the automated checks add commits to the branch that reset the tests. Then
we merge into main. (Submissions from forks are required to merge into a
non-main branch at first, since we don't have permissions to add commits to
branches in forks. I recently saw an example of how to work around this,
but that's the status quo workflow for now.)

bdeadman:
@skearnes thanks for the clarification. I thought it would be that. We will hold on that for now though. Me and @qai222 had a discussion at OH today, and there may be some changes we want to make to the dataset before pushing it to main.

qai222:
> @skearnes thanks for the clarification. I thought it would be that. We will hold on that for now though. Me and @qai222 had a discussion at OH today, and there may be some changes we want to make to the dataset before pushing it to main.

I uploaded my processing scripts to [this repo](https://github.com/qai222/ord_submission/blob/master/dataset_47463309/process.py). A few notes from the discussion with @bdeadman and the script:
1. The `solvent.volume` should have `volume_includes_solutes=True` specified. This field is unspecififed in the current submission. 
2. Convention needed for describing room temperature as a `Temperature` message.
3. Convention needed for describing pre-mix, e.g. 
   ```
   After placing the 96-well plates in ultrasonic water bath for 10 seconds to mix the reaction uniformly, 
   each 96-well plate was covered with an optical glass to minimize the volatilization of solvent and component
   ```
   (right now this is in `ReactionSetup.environment`).
4. The quantification of products includes a derivatization reaction. One can include it as `ReactionWorkups` or referencing another `Reaction` record. We chose the former. This led to ambiguity in assigning `reaction_role`: A is added as a catalyst of the derivatization reaction, should the role of A be `workup` or `catalyst`?
5. `ReactionInput` has `addition_order` specified, but `workups` is a sequence, so `workup1` precedes `workup2` implies `workup1.input.addition_order >= workup2.input.addition_order`. There should be a validation function for this.
6. Enantiomer products: what should the product SMILES be in reaction SMILES (one without chiral or two with chiral)? how should one describe the `ProductCompound`? What I did is
   ```
   ProductCompound(
    identifiers=[
        CompoundIdentifier(type="SMILES", value=product_s_smi),
        CompoundIdentifier(type="SMILES", value=product_r_smi),
    ],
    is_desired_product=True,
   ...
   ```

PR #192, opened 2024-07-17
submission 47465522

qai222:
reaction optimization plate from 10.1039/d3dd00167a

processing script: https://github.com/qai222/ord_submission/blob/master/dataset_47465522/process.py 

comments:
- pd catalysts were added as solutions and then the solvents were removed to leave solid solutes before adding other reactants/reagents.
- stirrers were added between reaction inputs
- ambiguous reaction role for copper cocatalyst
- some of the preparation were done in glove box, not sure where to put this info
- the vial was sealed during some of the processes, not sure where to put this info
  - one option is `VesselType.PRESSURE_FLASK` but this seems odd when vials are actually used
- no place to specify heater type (heated tumble stirrer)
- measurements are relative peak area against internal standard, normalized to 100%. I used `type=ProductMeasurement.ProductMeasurementType.AREA` to record them

### DIFF
--- a/dataset.pbtxt
+++ b/dataset.pbtxt
@@ -187,12 +187,25 @@ reactions {
   }
   conditions {
     temperature {
+      control {
+        type: CUSTOM
+        details: "heated tumble stirrer"
+      }
       setpoint {
         value: 100.0
         units: CELSIUS
       }
     }
+    pressure {
+      control {
+        type: SEALED
+        details: "Sealed 24-well metal heating blocks"
+      }
+    }
     conditions_are_dynamic: false
+  }
+  notes {
+    procedure_details: "\n        - reaction input `pd_catalyst` was added as a solution and then the solvent was removed to leave \n        solid solute before adding other reactants/reagents.\n        - stirrer was added after `pd_catalyst` and before `base`\n        - reaction inputs were prepared inside a glove box but the reaction itself did not happen in a glove box\n        - the reaction vial was sealed after adding all reaction inputs\n        "
   }
   workups {
     type: TEMPERATURE
@@ -369,7 +382,7 @@ reactions {
     doi: "10.1039/d3dd00167a"
     record_created {
       time {
-        value: "2024-07-17 03:07:17 UTC"
+        value: "2024-07-18 18:02:20 UTC"
       }
       person {
         username: "qai"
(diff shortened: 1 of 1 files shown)


skearnes:
> reaction optimization plate from 10.1039/d3dd00167a
> 
> processing script: https://github.com/qai222/ord_submission/blob/master/dataset_47465522/process.py
> 
> comments:
> 
> * pd catalysts were added as solutions and then the solvents were removed to leave solid solutes before adding other reactants/reagents.
> * stirrers were added between reaction inputs
> * ambiguous reaction role for copper cocatalyst
> * some of the preparation were done in glove box, not sure where to put this info

Maybe here? https://github.com/open-reaction-database/ord-schema/blob/main/proto/reaction.proto#L44

> * the vial was sealed during some of the processes, not sure where to put this info

Maybe here? https://github.com/open-reaction-database/ord-schema/blob/main/proto/reaction.proto#L778
   
>   * one option is `VesselType.PRESSURE_FLASK` but this seems odd when vials are actually used
> * no place to specify heater type (heated tumble stirrer)

Maybe here? https://github.com/open-reaction-database/ord-schema/blob/main/proto/reaction.proto#L574

> * measurements are relative peak area against internal standard, normalized to 100%. I used `type=ProductMeasurement.ProductMeasurementType.AREA` to record them

skearnes:
Other details that don't have an obvious home can go here: https://github.com/open-reaction-database/ord-schema/blob/main/proto/reaction.proto#L778

skearnes:
Might also consider this for the vessel sealing: https://github.com/open-reaction-database/ord-schema/blob/main/proto/reaction.proto#L601

skearnes:
> Might also consider this for the vessel sealing: https://github.com/open-reaction-database/ord-schema/blob/main/proto/reaction.proto#L601

And here: https://github.com/open-reaction-database/ord-schema/blob/main/proto/reaction.proto#L495

qai222:
The previous comments were solved by using fields of `ReactionNotes`  and `ReactionConditions`. 

One thing about the addition of pd catalyst. This `ReactionInput` was added as a solution and its solvent was removed using an evaporator before adding other `ReactionInput`:
1. what should be the `Texture` of this `ReactionInput`? I'm using `SOLID` as this is the form in which this `ReactionInput` was immediately before mixing with other `ReactionInput`.
   - in this dataset the author did not describe the addition device so I left it unspecified. There could be a situation where a liquid handler is described for a `ReactionInput` whose `texture` is `SOLID`, which could be confusing.
2. the removed solvent (and the concentration of the solution) could be useful but it's only included in an unstructured way, especially when this removed solvent is different from the solvent used in the actual reaction

FWIW Here is the related text from the paper
```
Inside a glovebox, a solution of each palladium precatalyst was prepared by dissolving 20 μmol of each
in a sufficient quantity of 2-methyltetrahydrofuran to give 2000 μL of solution. 100 μL of each solution
was then dosed to 8 x 24 well plates, and the solvent was removed in vacuo using the Genevac for 1 h.
The plates were stored in the glovebox and used within 1 year of creation.
```

bdeadman:
I'm reviewing this, but I will merge it into the branch already and we can finish the review there.

PR #206, opened 2024-10-04
Automated LC-MS analysis and data extraction for high-throughput chemistry

bdeadman:
I'm reviewing the dataset and will then replace the file before opening the PR. Will need to update the branch as well.


bdeadman:
Original dataset preparation by @qai222 in #192, and reviewed by me. As part of the review I have directly implemented my corrections into a replacement dataset file.

This is a 48 reaction dataset which was included in the paper introducing the PyParse LC processing software ([dx.doi.org/10.1039/d3dd00167a](dx.doi.org/10.1039/d3dd00167a)). In review I have made the following edits:

- Moved the details about how the catalyst input was added (as a solution, and then evaporated) to the addition device details with CUSTOM type.
- Replaced some special character codes (e.g. \316\274) with a simple text equivalent (e.g. mu -> u) to ensure it displays in ORD.
- Removed some line breaks and white space.
- Added CAS and MDL numbers for the catalysts - this is the start of the ORD catalyst library.
- Tweaked format of RecordCreation timestamp to ensure it validates in submission.

Preparation files for the revised dataset here [qai-2 BJD preparation files.zip](https://github.com/user-attachments/files/17653282/qai-2.BJD.preparation.files.zip)

bdeadman:
Oops - the new dataset may be missing the correct name and description. Fix incoming.

bdeadman:
@skearnes - another dataset for GitHub review. This was prepared by @qai222 but then I've made some changes in review. A full review isn't needed, but a quick check would be useful.

skearnes (approved):

PR #207, opened 2024-10-11
Aza-heck dataset submission

saratanov:
Submission of the Aza-Heck cyclisation dataset extracted from https://pubs.acs.org/doi/full/10.1021/acs.orglett.1c02767 as part of the ORD training program.


bdeadman:
@saratanov can you also please upload the template and spreadsheet into the chat here. You will need to zip them into a file before GitHub will accept them as an upload.

saratanov:
Here are the templates and spreadsheets - one has NMR yield data and the other does not.
[aza_heck.zip](https://github.com/user-attachments/files/17366215/aza_heck.zip)

bdeadman:
@saratanov sorry for the delay. Overall it is very good. Some things to check and or update below. Happy to go through this in a call if it helps.

- [ ]  The sample number (e.g. A1, D6) should be included as a reaction identifier. This conveys information about where the rxn was on the plate.
- [ ]  Check the that there aren't complex characters in the record. The pbtxt file has "\342\200\213" in the catalyst name of the first reaction, but no symbol is displayed in the editor. Other placeholders (e.g. \316\274 for micro, u) are also in the file. Please replace these tags with a text label (e.g A, B, u etc). This may be easiest to do with find/replace in a text editor.
- [ ]  The addition device is currently "CUSTOM" with "micropipettor" in details. Change the type to PIPETTE and the details field is no longer required. How was the ligand input added? Probably weighing in. If so the addition device can be CUSTOM with "manual weighing" in the details field. Adding a specific type for weighing is on the list to do.
- [ ]  The vessel preparation message for putting the teflon sheet and mats could be changed to a vessel attachment of type MAT with the text in the details field.
- [ ] Remove the stir bars vessel attachment. It is described in the stirring conditions, and it would be consistent with existing records to not include it in the attachments list.
- [ ] Conditions - it sounds like the addition of IS should be moved to the workups (from the inputs) and a "cool down to RT" workup also added. Then the temperature of the reaction would be a static 80C.
- [ ]  Make the temperature control type "DRY_ALUMINUM_PLATE" [excuse the US spelling!].
- [ ]  The stirring workup seems odd. From the details it might be better as an ALIQUOT step, and a separate STIRRING step.
- [ ]  Include a time value in the Outcome which is how long it has reacted for at the point the measurement is taken, typically excluding workup (24 hours?).
- [ ]  For the analysis we need to create the Analysis options (+ add analysis underOutcome). The label given here (like in the input headings) is then the 'key' you select in the "based on" drop down menus under the analysis measurements.
- [ ]  Provenance - we can assume that Katerina Korch did the experiment at Uni of Delaware.
- [ ]  I would also recommend that you set up your ORCiD account, and include the number on this submission. You will want this when you start publishing your PhD work.

bdeadman:
Reaction 23 also seems to have 2 outcomes. Multiple analyses and products are grouped into one outcome if they have the same reaction time.

saratanov:
[aza_heck_v2.zip](https://github.com/user-attachments/files/18026495/aza_heck_v2.zip)

@bdeadman Apologies for the delay, I have implemented the changes you suggested.

bdeadman:
Thanks @saratanov. I'll review these early next week and hopefully we will push it through soon after.

bdeadman:
- [ ] Input key 'Catalyst/n'

bdeadman:
Replacement files here: [aza_heck_v2.1.zip](https://github.com/user-attachments/files/18471677/aza_heck_v2.1.zip)

Thanks for making the changes @saratanov. I've reviewed it and made a few more edits directly. See above comments (no action needed from you), and let me know if you want an explanation for any of them.

From here I will approve this PR (even though it is the wrong file) to get it into our branch where I will be able to swap in the new files and (assuming everything still validates) merge it into the public database.

PR #212, opened 2024-11-16
Doyle group Chan-Lam coupling dataset

bdeadman:
Adding Doyle group Chan Lam coupling dataset on behalf of Shivaani. Further details to be added.

### DIFF
--- a/data/Doyle_Group_Chan_Lam_Dataset_5.pb.gz
+++ b/data/Doyle_Group_Chan_Lam_Dataset_5.pb.gz
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:f934b7114f5e4aa27d38532f5df1dbad272dc6db663f24ea4de4c5e08e230f00
+size 581559


skearnes:
@bdeadman do you have the supporting notebook for this? I'm trying to resolve the date format; see https://github.com/open-reaction-database/ord-logbook/blob/main/entries/2026-09-03-date-time-formats/README.md

bdeadman:
I think the attached folder is the latest version of the files used to prepare the dataset. 

[submission.zip](https://github.com/user-attachments/files/31832348/submission.zip)

This one was also built by template enumeration over a csv, but the enumeration script is in a notebook. The residual time stamps in the notebook could possibly be relevant to solving the problem, but otherwise I would assume the dataset's record creation dates should be in Nov 2024.

Note the original template has an ambiguous record creation date of "08/05/2024, 11:34:02" which I think will be dd/mm/yyyy format going by my OS timestamps on the very first versions of the template.

bdeadman:
Confirmed that the record creation dates in the template are 8 May 2024 by going through my correspondence with Shivaani. Although the template was originally built by Shivaani in the US (and would likely be mm/dd), I had to rebuild the template in May 2024 (using the legacy editor) to overcome some undiagnosed problem that blocked the enumeration. I would have used  dd/mm format in preference to mm/dd.

PR #214, opened 2024-12-18
Replace dataset with correction

bdeadman:
Correction to Golden dataset.


bdeadman:
@spencerheins this will be the update to your dataset. After considering it some more I realised that my previous update proposal would have replaced all the reaction ids. This version has been done in such a way to maintain the ids.

I've documented the process in a new [example Jupyter Notebook](https://github.com/open-reaction-database/ord-schema/blob/ff508db8bb67c5f44d1f5a00a3dfe096fcd5dea0/examples/updates/1_Golden/update-dataset.ipynb) that I would like to include in ord-schema open-reaction-database/ord-schema#758. Let me know if you have any issues with it being used as an example. 

Also note that I have attributed the record_modification to myself for two reasons 1) if there is a problem with the update they will come to me, and 2) I wanted to include some example person information in the example notebook.

spencerheins:
@bdeadman - I've got no issue with using this as an example in the JN. Thanks again for the assistance.

bdeadman:
@connorcoley @skearnes - this is a correction to the Golden dataset from Wisconsin Madison. The update is documented in a Jupyter notebook (link in earlier comment) and involved updating some SMILES strings, and adding a record modified event.

connorcoley (approved):

PR #215, opened 2025-01-16
Correction to dataset description

bdeadman:
Remove 'GSK', and correct 'Janssen' spelling in dataset_description. Note that  no record_modified provenance added.


bdeadman:
@skearnes this is the recently added dataset from the Doyle group. I've corrected a couple of mistakes in the dataset description. Since it is only the dataset description changing, I haven't added a 'record_modified' event to the reactions.

skearnes (approved):

PR #282, opened 2026-08-03
Pick up ord-schema 0.8.2, validate ids, and let a dependency bump test itself

skearnes:
## Summary

Moves the `ord-schema` pin from `>=0.8` to `>=0.8.2`, has both validation jobs pass the new `--validate_ids` flag, and fixes the guard that stopped a PR like this one from testing itself.

0.8.2 carries the validation speedups — identifier parsing memoized, and findings collected through an explicit context instead of the `warnings` module — together **2.84×** on the uspto file, which is the long pole of every full-corpus run (~75 min of the last one).

## Changes

- `ord-schema>=0.8.2` in `pyproject.toml`, `uv.lock` relocked.
- Both `validate_pb` and `validate_parquet` pass `--validate_ids`. Submissions are validated with `validate_ids=True` in `process_dataset.py`, so ids are well-formed on the way in, but the sweep ran with the defaults and never looked at them again — an id that stopped being valid after submission would have passed forever. The flag is opt-in upstream because ids are assigned *during* submission, so a draft does not have them yet.
- **The `pipeline/` checkout now comes from `github.sha` except on fork PRs.** It was pinned to the base branch on every PR, but the workflow file itself comes from the merge ref — so the first push of this PR ran main's `uv.lock` (ord-schema 0.8.0) against its own command line and every shard died on `unrecognized arguments: --validate_ids`. The guard only ever needed to stop a **fork** from choosing the environment its own data is validated in; anyone who can open a same-repo PR can already edit this file.
- CLAUDE.md records the flag, and why this guard is deliberately narrower than `submission.yml`'s.

## Testing

`pytest -n auto` (52 passed), `ruff check`, `ty`, and `pre-commit` all clean against 0.8.2 — that release's API changes (`ValidationError` is now an `Exception`, `ValidationWarning` removed, leaf validators private) touch nothing this repository uses.

Ran the exact command CI runs — `python -m ord_schema.scripts.validate_dataset --input_pattern="data/*/*.parquet" --validate_ids --n_jobs=4` — over the whole local parquet corpus under 0.8.2: 53 datasets, 2466 row groups, exit 0. Beforehand I checked that all 552 corpus dataset ids and 13,108 sampled reaction ids are well-formed, so the flag is a no-op today and a guard from here on.

This PR touches `validation.yml`, so its own run is the full 11-shard matrix — which is also the first real test of `--validate_ids` against the 550 `.pb.gz` datasets I could not materialize locally.

## Expect more warnings, not more errors

The sweep will report roughly **315 new "leading or trailing whitespace" warnings** across the corpus. Those come from [ord-schema#917](https://github.com/open-reaction-database/ord-schema/pull/917), which landed between 0.8.0 and 0.8.1 — not from the validation refactors. I compared 54 datasets across four builds (pinned 0.8.0, the commit before the refactors, `main`, and the refactor branch): the refactors change **no findings at all**, once #927's deliberate reformat of the inconsistent-identifiers message is normalized. They are warnings, so validation still passes.

<details><summary><h3>Greptile Summary</h3></summary>

The PR upgrades the validation pipeline to `ord-schema` 0.8.2 and enables identifier validation during both protobuf and Parquet corpus sweeps.
- Passes `--validate_ids` in both validation jobs.
- Uses the merge commit’s dependency environment for same-repository pull requests while preserving the base-branch environment for fork pull requests.
- Updates the dependency declaration, lockfile, and workflow documentation.
</details>

<h3>Confidence Score: 5/5</h3>

The pull request appears safe to merge.

No blocking failure remains.

<details open><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| .github/workflows/validation.yml | Enables identifier validation and correctly aligns same-repository PR commands with their locked schema environment while retaining the fork trust boundary. |
| pyproject.toml | Raises the pipeline dependency floor to ord-schema 0.8.2. |
| uv.lock | Relocks ord-schema at 0.8.2 with consistent dependency metadata and package hashes. |
| CLAUDE.md | Documents identifier validation and the workflow’s fork-aware pipeline checkout behavior. |

</details>

<sub>Reviews (3): Last reviewed commit: ["Distill the comments this branch added"](https://github.com/open-reaction-database/ord-data/commit/d6b53bc06fec81d8982ed4e93c67c89e4d108fa8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49542629)</sub>

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -93,14 +93,18 @@ jobs:
       run: |
         git config lfs.url "https://github.com/${GITHUB_REPOSITORY}.git/info/lfs"
         git lfs pull --include="${FILTER}/*.pb*"
-    # A pull_request run checks out the merge ref, so the tree above is
-    # contributor-controlled, uv.lock included. The environment is built from a
-    # ref this repository owns instead: the base branch on a pull request, the
-    # pushed or scheduled ref otherwise.
-    - name: Check out the pipeline from a trusted ref
+    # A fork's pull request must not choose the environment its own data is
+    # validated in, so a fork run builds from the base branch. Every other run
+    # builds from the ref under test -- for a same-repo pull request that is the
+    # merge ref, so a change to uv.lock and a change to how the validator is
+    # invoked can land together and actually be tested. Anyone who can open a
+    # same-repo pull request can already edit this file.
+    - name: Check out the pipeline
       uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       with:
-        ref: ${{ github.event.pull_request.base.ref || github.ref }}
+        ref: >-
+          ${{ github.event.pull_request.head.repo.fork
+              && github.event.pull_request.base.ref || github.sha }}
         path: pipeline
         lfs: false
         # Only what `uv sync` reads. Leaving out data/ also keeps a second tree
@@ -166,12 +170,14 @@ jobs:
         else
           git lfs pull --include="${LFS_INCLUDE}"
         fi
-    # See validate_pb: the environment comes from a ref this repository owns,
-    # never the contributor-controlled merge ref.
-    - name: Check out the pipeline from a trusted ref
+    # See validate_pb: forks build from the base branch, everything else from
+    # the ref under test.
+    - name: Check out the pipeline
       uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       with:
-        ref: ${{ github.event.pull_request.base.ref || github.ref }}
+        ref: >-
+          ${{ github.event.pull_request.head.repo.fork
+              && github.event.pull_request.base.ref || github.sha }}
         path: pipeline
         lfs: false
         # Only what `uv sync` reads. Leaving out data/ also keeps a second tree
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -48,7 +48,9 @@ concurrency:
   cancel-in-progress: true
 
 # validate_dataset.py ships inside the ord_schema wheel, so both jobs below run
-# it as a module out of the `pipeline` dependency group. uv.lock is therefore
+# it as a module out of the `pipeline` dependency group. Both pass --validate_ids,
+# matching what process_dataset.py checks when a dataset is submitted, so an id
+# that stopped being well-formed after submission does not go unnoticed. uv.lock is therefore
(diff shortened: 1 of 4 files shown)


skearnes:
## Validation times, before and after

All 11 shards green. Times are the `Validate …` step only — LFS fetch and environment setup excluded — so this isolates the library change. Baseline is #280's run, which used ord-schema 0.8.0 through the same lockfile path.

| shard | 0.8.0 | 0.8.2 | speedup |
|---|---:|---:|---:|
| **uspto** | 81.8 min | 17.6 min | 4.65x |
| `data/[0-4][5-9]` | 15.8 min | 5.0 min | 3.15x |
| `data/[a-f][a-f]` | 14.3 min | 4.4 min | 3.25x |
| other parquet | 11.7 min | 4.3 min | 2.71x |
| `data/[0-4][a-f]` | 5.4 min | 3.8 min | 1.44x |
| `data/[a-f][5-9]` | 12.6 min | 3.5 min | 3.60x |
| `data/[5-9][a-f]` | 9.4 min | 3.5 min | 2.72x |
| `data/[5-9][0-4]` | 10.6 min | 3.3 min | 3.23x |
| `data/[a-f][0-4]` | 9.5 min | 3.0 min | 3.20x |
| `data/[0-4][0-4]` | 6.8 min | 2.2 min | 3.07x |
| `data/[5-9][5-9]` | 6.4 min | 1.5 min | 4.29x |
| **total (core-min)** | **184.5** | **52.1** | **3.54x** |

**Wall clock for the matrix is set by the longest shard: 81.8 min → 17.6 min.**

Two reasons to read 3.54× as a floor rather than a headline:

- This run does *more* work than the baseline. It passes `--validate_ids`, which checks every dataset and reaction id, and 0.8.2 also carries ord-schema#917's whitespace and CXSMILES checks — the ~315 new warnings noted in the PR description. The speedup is net of both.
- The corpus is byte-identical between the two runs, and the shard definitions are unchanged, so nothing else moved.

It also beats what I measured locally (2.84×, from 1.785 → 0.625 ms/reaction). I'd rather not assert a mechanism I haven't verified — runner cores differ from this laptop's in ways that would change the balance between the native RDKit work the memoization removes and the Python around it — but the local figure was evidently the conservative one.

`data/[0-4][a-f]` is the outlier at 1.44×. It was already the fastest pb shard at 5.4 min, so it has the least chemistry to skip; the memoization only pays where identifier parsing dominates.

_Analysis performed with [Claude Code](https://claude.com/claude-code) (Claude Opus 5)._

